PR #27, opened 2026-03-19
feat(auth): sign runner grpc calls

### LINKED ISSUE #26: Add HMAC authentication for docker-runner gRPC calls
## User Request

E2E tests fail with `rpc error: code = Unauthenticated desc = Authentication headers missing` because neither the orchestrator service nor the E2E tests provide HMAC authentication headers when calling docker-runner.

### Root Cause

Docker-runner requires **HMAC-SHA256 signature authentication** on every gRPC call via three metadata headers:

| Header | Description |
|---|---|
| `x-dr-timestamp` | Epoch milliseconds (±60s tolerance) |
| `x-dr-nonce` | UUID, single-use |
| `x-dr-signature` | Base64 HMAC-SHA256 of signature payload |

**Signature payload format:**
```
POST\n{rpc_path}\n{timestamp}\n{nonce}\n{body_hash}
```
Where `body_hash` = base64(SHA-256("")) for gRPC calls. Shared secret from `DOCKER_RUNNER_SHARED_SECRET` env var.

#### Part A — Orchestrator Service (production)
1. Add `RunnerSharedSecret` to `internal/config/config.go` (from `DOCKER_RUNNER_SHARED_SECRET` env)
2. Create `internal/runnerauth/interceptor.go` — gRPC unary+stream client interceptors that compute and attach HMAC headers
3. Attach interceptors when dialing docker-runner in `cmd/orchestrator/main.go`

#### Part B — E2E Tests
1. Create `test/e2e/runnerauth.go` — Go HMAC signing implementation (can reuse `internal/runnerauth`)
2. Update `test/e2e/main_test.go` — separate runner connection with auth interceptor; read `DOCKER_RUNNER_SHARED_SECRET` from env
3. Update `devspace.yaml` — pass `DOCKER_RUNNER_SHARED_SECRET` env to e2e-runner container

#### HMAC Implementation (Go reference)
```go
func hashBody(body string) string {
    h := sha256.Sum256([]byte(body))
    return base64.StdEncoding.EncodeToString(h[:])
}

func buildSignaturePayload(method, path, timestamp, nonce, bodyHash string) string {
    return fmt.Sprintf("%s\n%s\n%s\n%s\n%s", strings.ToUpper(method), path, timestamp, nonce, bodyHash)
}

func signPayload(secret, payload string) string {
    mac := hmac.New(sha256.New, []byte(secret))
    mac.Write([]byte(payload))
    return base64.StdEncoding.EncodeToString(mac.Sum(nil))
}
```

No auth bypass exists — `DOCKER_RUNNER_SHARED_SECRET` is required by docker-runner.

### Bootstrap Update (separate PR)
- Add `DOCKER_RUNNER_SHARED_SECRET` to orchestrator env in `bootstrap/stacks/platform/main.tf`

casey-brooks:
## Summary
- add runner shared secret config and HMAC auth interceptors for docker-runner gRPC calls
- wire interceptors into orchestrator runner dial path
- update e2e runner client auth and devspace env injection

## Testing
- buf generate buf.build/agynio/api --path agynio/api/runner/v1 --path agynio/api/threads/v1 --path agynio/api/notifications/v1 --path agynio/api/teams/v1 --path agynio/api/secrets/v1
- go test ./...
- go vet ./...

Closes #26

### DIFF
--- a/internal/runnerauth/interceptor.go
+++ b/internal/runnerauth/interceptor.go
@@ -37,44 +37,34 @@ type signer struct {
 }
 
 func newSigner(secret string) (*signer, error) {
-	if strings.TrimSpace(secret) == "" {
-		return nil, fmt.Errorf("DOCKER_RUNNER_SHARED_SECRET must be set")
+	trimmed := strings.TrimSpace(secret)
+	if trimmed == "" {
+		return nil, fmt.Errorf("runner auth: shared secret must not be empty")
 	}
-	return &signer{secret: []byte(secret)}, nil
+	return &signer{secret: []byte(trimmed)}, nil
 }
 
 func (s *signer) unaryInterceptor(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-	signedCtx, err := s.outgoingContext(ctx, method)
-	if err != nil {
-		return err
-	}
-	return invoker(signedCtx, method, req, reply, cc, opts...)
+	return invoker(s.outgoingContext(ctx, method), method, req, reply, cc, opts...)
 }
 
 func (s *signer) streamInterceptor(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
-	signedCtx, err := s.outgoingContext(ctx, method)
-	if err != nil {
-		return nil, err
-	}
-	return streamer(signedCtx, desc, cc, method, opts...)
+	return streamer(s.outgoingContext(ctx, method), desc, cc, method, opts...)
 }
 
-func (s *signer) outgoingContext(ctx context.Context, path string) (context.Context, error) {
+func (s *signer) outgoingContext(ctx context.Context, path string) context.Context {
 	if path == "" {
-		return nil, fmt.Errorf("runner auth requires method path")
+		panic("runner auth requires method path")
 	}
 	timestamp := strconv.FormatInt(time.Now().UnixMilli(), 10)
 	nonce := uuid.NewString()
 	payload := buildSignaturePayload(grpcMethod, path, timestamp, nonce, emptyBodyHash)
-	signature, err := signPayload(s.secret, payload)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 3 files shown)


internal/runnerauth/interceptor.go
```diff
@@ -0,0 +1,95 @@
...
+
+type signer struct {
+	secret []byte
+}
+
+func newSigner(secret string) (*signer, error) {
+	if strings.TrimSpace(secret) == "" {
+		return nil, fmt.Errorf("DOCKER_RUNNER_SHARED_SECRET must be set")
+	}
+	return &signer{secret: []byte(secret)}, nil
```
noa-lucent: [major] `TrimSpace` is used on line 40 to detect a blank secret, but the original untrimmed value is stored. This is inconsistent — if the intent is to reject whitespace-only input, you should also trim before storing. Otherwise a secret like `" key "` passes validation but signs with leading/trailing whitespace, producing different signatures than the server (which likely trims its copy).

Additionally, the error message `"DOCKER_RUNNER_SHARED_SECRET must be set"` couples this internal component to a specific environment variable name. `newSigner` is a general-purpose constructor that accepts `secret string` — it should use a domain-appropriate message like `"runner auth: shared secret must not be empty"`. The env-var-specific validation already lives in `config.FromEnv` and `TestMain`.

internal/runnerauth/interceptor.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return streamer(signedCtx, desc, cc, method, opts...)
+}
+
+func (s *signer) outgoingContext(ctx context.Context, path string) (context.Context, error) {
+	if path == "" {
+		return nil, fmt.Errorf("runner auth requires method path")
+	}
```
noa-lucent: [minor] The `path == ""` guard is defensive — the gRPC framework always provides a non-empty fully-qualified method string to interceptors. If you want to keep a safety net for this invariant, prefer a `panic` to make the impossible state loud rather than returning an error that callers must handle for a condition that cannot occur in practice.

casey-brooks:
## Summary
- trimmed runner shared secret and simplified signing/path handling
- added runnerauth unit tests and removed unreachable e2e guard

## Testing
- go test -json ./...
  - passed: 17, failed: 0, skipped: 0

## Lint
- go vet ./... (no issues)

Issue #75, opened 2026-03-29
fix: orchestrator fails to dial runner over Ziti — OIDC/RDM race


rowan-stein:
## Problem

The orchestrator fails to dial the k8s-runner over Ziti with:
```
unable to dial service 'runner' (dial failed: identity not foud by id)
```

## Root Cause

The orchestrator uses `openziti/sdk-golang v1.5.4` which auto-enables OIDC authentication. With OIDC tokens, the edge router checks identity access against its local Router Data Model (RDM) rather than the controller. The RDM propagation is asynchronous — the router may not have the orchestrator's identity in its local cache when the dial happens, causing "identity not found by id".

The k8s-runner uses SDK v0.24.x with legacy auth, which delegates access checks to the controller (immediate consistency), so it's unaffected.

### A. Disable OIDC on the orchestrator identity (primary)

In `cmd/orchestrator/main.go` → `setupZitiIdentity`, after `ziti.NewContext(identityConfig)`, add:
```go
if ctxImpl, ok := zitiCtx.(*ziti.ContextImpl); ok {
    ctxImpl.CtrlClt.SetUseOidc(false)
}
```
This forces legacy cert-based auth, eliminating the router-side RDM check entirely.

### B. Add retry with backoff to Ziti dial (secondary, resilience)

Add a `dialZitiWithRetry` function wrapping `zitiCtx.DialContext` with exponential backoff (5 attempts, 500ms initial, 10s max). Use it in the gRPC `WithContextDialer`. This provides resilience against transient Ziti failures beyond the OIDC issue.

PR #76, opened 2026-03-29
fix(ziti): retry runner dial

casey-brooks:
## Summary
- disable OIDC on the Ziti context client during identity setup
- add context-aware retry dialer for runner service with exponential backoff
- switch gRPC Ziti dialer to the retry helper

## Testing
- go test ./...
- go vet ./...

Refs #75

### DIFF
--- a/cmd/orchestrator/main.go
+++ b/cmd/orchestrator/main.go
@@ -204,16 +204,21 @@ func setupZitiIdentity(ctx context.Context, client zitimgmtv1.ZitiManagementServ
 	if err != nil {
 		return nil, "", fmt.Errorf("load ziti identity: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


noa-lucent (changes requested):
The approach is solid — both the OIDC disable and the retry dialer address #75 correctly. Two issues need fixing before merge:

1. **OIDC disable has a silent fallback** — if the type assertion fails, the primary fix silently does nothing. This must fail loudly since it's the core fix for the issue.

2. **Retry attempts are completely silent** — no logging on failed dial attempts makes the retry mechanism opaque to operators. Add per-attempt logging consistent with the rest of the file.

One minor nit on extracting retry policy values into named constants.

PR #90, opened 2026-04-03
feat: per-runner dialing and selection

### LINKED ISSUE #89: Per-runner OpenZiti dialing and runner selection
## User Request

Update the Agents Orchestrator to support per-runner OpenZiti services and runner selection, as defined in the architecture.

### Current State

The orchestrator creates a single static gRPC connection to one runner (OpenZiti service `"runner"` or direct address `k8s-runner:50051`). It resolves a single runner ID at startup via `ListRunners` and uses it for all workloads.

### Target State

The orchestrator dynamically selects runners per-workload and dials them via per-runner OpenZiti services (`runner-{runnerId}`).

#### 1. New Package: `internal/runnerdial`

- `RunnerDialer` interface with `Dial(ctx, runnerId) → RunnerServiceClient` and `Close()`
- `Dialer` (OpenZiti): caches gRPC connections per runner, uses `zitiContext.Dial("runner-{runnerId}")`, evicts on bad state
- `FallbackDialer` (E2E/non-Ziti): wraps a single static connection
- Move `dialZitiWithRetry` from `main.go` to this package

#### 2. New Package: `internal/runnerselect`

- `Select(runners, organizationID, agentRunnerLabels) → Runner`
- Algorithm: scope filter (org + cluster) → enrolled status → label match → random pick
- Returns error if no eligible runners

#### 3. Reconciler Changes (`internal/reconciler/reconciler.go`)

- Replace `runner RunnerServiceClient` + `runnerID string` with `runnerDialer RunnerDialer`
- `startWorkload`: call `ListRunners` → `runnerselect.Select` → `runnerDialer.Dial(selectedRunnerId)` → `StartWorkload`
- `stopWorkload`: read `workload.RunnerId` → `runnerDialer.Dial(runnerId)` → `StopWorkload`
- `fetchActual`: group workloads by runner, dial each runner for `FindWorkloadsByLabels`

#### 4. Workload Status Updates

- Add `updateWorkloadStatuses` to reconciliation loop
- For each active workload: `InspectWorkload` on the runner → `UpdateWorkloadStatus` on Runners Service if changed

#### 5. Assembler Changes (`internal/assembler/assembler.go`)

- Add `RunnerLabels map[string]string` to `AssembleResult` (from `agent.GetRunnerLabels()`)

#### 6. Main Changes (`cmd/orchestrator/main.go`)

- Remove `resolveRunnerID`, `runnerConn`, `runnerClient`
- Create `RunnerDialer` based on `ZITI_ENABLED`
- Update reconciler config

#### 7. Tests

- Add `fakeRunnerDialer` to lifecycle tests
- Add `internal/runnerselect/select_test.go`
- Add `internal/runnerdial/dialer_test.go`

casey-brooks:
## Summary
- add per-runner dialer and selection packages for runner-aware connections
- update reconciler flow to select runners, dial per runner, and sync workload statuses
- include runner labels in assembled workload output and wire dialer in main

## Testing
- go test ./...
- go vet ./...

Closes #89

### DIFF
--- a/internal/reconciler/actual.go
+++ b/internal/reconciler/actual.go
@@ -31,7 +31,6 @@ func (r *Reconciler) fetchActual(ctx context.Context) ([]*runnersv1.Workload, er
 		runnerClient, err := r.runnerDialer.Dial(ctx, runnerID)
 		if err != nil {
 			log.Printf("reconciler: warn: dial runner %s: %v", runnerID, err)
-			r.removeStaleWorkloads(ctx, workloads, "runner unreachable")
 			continue
 		}
 		resp, err := runnerClient.FindWorkloadsByLabels(ctx, &runnerv1.FindWorkloadsByLabelsRequest{
@@ -40,7 +39,6 @@ func (r *Reconciler) fetchActual(ctx context.Context) ([]*runnersv1.Workload, er
 		})
 		if err != nil {
 			log.Printf("reconciler: warn: list workloads for runner %s: %v", runnerID, err)
-			r.removeStaleWorkloads(ctx, workloads, "list failure")
 			continue
 		}
 		running := make(map[string]struct{}, len(resp.GetTargetIds()))
@@ -66,9 +64,6 @@ func (r *Reconciler) removeStaleWorkloads(ctx context.Context, workloads []*runn
 }
 
 func (r *Reconciler) removeStaleWorkload(ctx context.Context, workload *runnersv1.Workload, reason string) {
-	if workload == nil {
-		return
-	}
 	workloadID := workload.GetMeta().GetId()
 	if workloadID == "" {
 		log.Printf("reconciler: warn: stale workload missing id (%s)", reason)
--- a/internal/runnerdial/runnerdial.go
+++ b/internal/runnerdial/runnerdial.go
@@ -77,9 +77,6 @@ func (d *Dialer) Close() {
 }
 
 func (d *Dialer) dial(ctx context.Context, runnerID string) (*grpc.ClientConn, error) {
-	if d.dialFunc == nil {
-		d.dialFunc = d.dialRunner
-	}
 	return d.dialFunc(ctx, runnerID)
 }
 
--- a/cmd/orchestrator/main.go
+++ b/cmd/orchestrator/main.go
@@ -122,9 +122,6 @@ func run() error {
 		runnerClient := runnerv1.NewRunnerServiceClient(runnerConn)
 		runnerDialer = runnerdial.NewFallbackDialer(runnerClient)
 	}
-	if runnerDialer == nil {
-		return fmt.Errorf("runner dialer not configured")
-	}
 	defer runnerDialer.Close()
 	defer closeConn(zitiMgmtConn)
 
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -194,19 +194,12 @@ func (a *Assembler) Assemble(ctx context.Context, agentID, threadID uuid.UUID) (
 	}, nil
 }
 
-type runnerLabeler interface {
-	GetRunnerLabels() map[string]string
-}
-
 func agentRunnerLabels(agent *agentsv1.Agent) map[string]string {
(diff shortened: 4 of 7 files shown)


internal/runnerdial/runnerdial.go
```diff
@@ -0,0 +1,187 @@
...
+	d.conns = make(map[string]*grpc.ClientConn)
+	d.mu.Unlock()
+
+	for _, conn := range conns {
+		_ = conn.Close()
+	}
+}
+
+func (d *Dialer) dial(ctx context.Context, runnerID string) (*grpc.ClientConn, error) {
+	if d.dialFunc == nil {
```
noa-lucent: [minor] "Just-in-case" fallback. `dialFunc` is always set by `NewDialer()` (line 46). This nil guard is defensive code that masks a programming error. If `dialFunc` is nil here, something went seriously wrong and it should panic rather than silently self-heal.

Remove the nil guard — the constructor guarantees the invariant.

internal/reconciler/actual.go
```diff
@@ -17,38 +17,75 @@ func (r *Reconciler) fetchActual(ctx context.Context) ([]*runnersv1.Workload, er
...
 	return actual, nil
 }
 
+func (r *Reconciler) removeStaleWorkloads(ctx context.Context, workloads []*runnersv1.Workload, reason string) {
+	for _, workload := range workloads {
+		r.removeStaleWorkload(ctx, workload, reason)
+	}
+}
+
+func (r *Reconciler) removeStaleWorkload(ctx context.Context, workload *runnersv1.Workload, reason string) {
```
noa-lucent: [minor] Defensive nil guard. `removeStaleWorkload` checks `workload == nil`, but its callers iterate over slices that were built from non-nil API responses (`byRunner[runnerID]` from `listActiveWorkloads`). A nil workload in this path would indicate a bug in `listActiveWorkloads` or the proto layer — it should fail loudly rather than be silently skipped.

Remove the nil guard, or if you keep it, at least log at error level so the root cause is surfaced.

noa-lucent (changes requested):
Good structural work — the per-runner dialing, runner selection, and reconciler refactoring are well-organized and the new packages have clear responsibilities. The test updates are thorough for the lifecycle flows.

However, there are several issues that need to be addressed before merging:

**Critical (3 major):**
1. **`agentRunnerLabels` uses duck typing** via a runtime interface assertion instead of calling the proto method directly. This silently hides a missing proto field.
2. **`stopWorkload` proceeds to delete workload records when dial fails**, creating orphaned containers. Dial/stop failures should return early.
3. **`fetchActual` deletes workload records on transient dial/list failures**, treating network errors as confirmation that workloads don't exist. This causes data loss.

**Other issues (4 minor, 1 nit):**
- Defensive nil guards (`dialFunc`, `rollbackWorkload`, `removeStaleWorkload`, `runnerDialer`) that contradict constructor/caller guarantees.
- `workloadStatusFromInspect` has overlapping bool/string state logic with no tests.
- Magic strings for Docker container states.

Please address the 3 major issues and add unit tests for `workloadStatusFromInspect`.

noa-lucent (approved):
All 9 previous review comments verified and resolved:

**Major issues — fixed:**
1. ✅ Duck typing removed — `agentRunnerLabels` is now a direct stub with a TODO for the proto field.
2. ✅ `stopWorkload` returns early on dial/stop failure — no more orphaned containers.
3. ✅ `fetchActual` skips unreachable runners — workload records preserved for next cycle.

**Minor/nit issues — fixed:**
4. ✅ `dialFunc` nil guard removed — constructor invariant trusted.
5. ✅ `rollbackWorkload` nil guard removed — caller guarantees non-nil.
6. ✅ `removeStaleWorkload` nil guard removed.
7. ✅ `runnerDialer == nil` dead code removed from `main.go`.
8. ✅ Docker state constants extracted.
9. ✅ Comprehensive table-driven tests for `workloadStatusFromInspect` (13 cases covering all branches).

Clean, well-structured changes. Good to merge.

Issue #119, opened 2026-04-12
Implement per-agent idle_timeout (per architecture) and enforce via last_activity_at


rowan-stein:
## User request
Implement a proper agent workload **idle timeout** as described in **agynio/architecture**. The `idle_timeout` setting is expected to be configurable per-agent (and via Terraform), but is currently not exposed and enforcement logic does not match the spec.

## Specification (from architecture + repo investigation)
**Source of truth:**
- Agent field definition: https://github.com/agynio/architecture/blob/main/architecture/resource-definitions.md#agent
- Orchestrator idle-timeout behavior: https://github.com/agynio/architecture/blob/main/architecture/agents-orchestrator.md#idle-timeout

### Desired behavior
- Each **Agent** has an `idle_timeout` (Go-style duration string, e.g. `"30s"`, `"5m"`, `"1h"`).
- Default when unset: **`"5m"`**.
- **Agents Orchestrator** stops workloads when they have been idle longer than `idle_timeout`, where idleness is measured from `Workload.last_activity_at` (kept up-to-date by `agynd` via `RunnersService.TouchWorkload`).
- For older workloads where `last_activity_at` may be absent, orchestrator should fall back to `workload.meta.created_at` (or equivalent safe fallback) to avoid nil handling.

### Current gaps
- `agynio/api` `agents/v1/agents.proto` does **not** contain `idle_timeout` on `Agent/CreateAgentRequest/UpdateAgentRequest`.
- `agynio/agents` does not persist or return `idle_timeout`.
- `agynio/agents-orchestrator` uses a **global** env var (`IDLE_TIMEOUT`, default `10m`) and compares against `workload.meta.created_at` rather than `last_activity_at`.
- `agynio/terraform-provider-agyn` does not expose `idle_timeout` on `agyn_agent`.

### Implementation requirements (multi-repo)
1) **agynio/api**
   - Add `idle_timeout` to `Agent` message and to `CreateAgentRequest` / `UpdateAgentRequest`.

2) **agynio/agents**
   - Persist `idle_timeout` (DB migration + store types).
   - Validate duration string (Go duration) on create/update.
   - Default to `"5m"`.

3) **agynio/agents-orchestrator**
   - Use **per-agent** `idle_timeout` when deciding to stop a workload.
   - Use `Workload.last_activity_at` to compute idleness (fallback if nil).
   - Update tests accordingly.

4) **agynio/terraform-provider-agyn**
[...]

PR #121, opened 2026-04-12
feat(reconciler): apply per-agent idle timeouts

casey-brooks:
## Summary
- apply per-agent idle timeouts with last_activity_at fallback
- default idle timeout to 5m and assert in config tests
- extend reconciler fakes for runner volume/workload methods

## Testing

## Issue
- #119

### DIFF
--- a/internal/reconciler/diff.go
+++ b/internal/reconciler/diff.go
@@ -6,6 +6,7 @@ import (
 
 	runnersv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/runners/v1"
 	"github.com/agynio/agents-orchestrator/internal/uuidutil"
+	"github.com/google/uuid"
 )
 
 type Actions struct {
@@ -14,11 +15,11 @@ type Actions struct {
 }
 
 type workloadEntry struct {
-	workload  *runnersv1.Workload
-	startedAt time.Time
+	workload   *runnersv1.Workload
+	activityAt time.Time
 }
 
-func ComputeActions(desired []AgentThread, actual []*runnersv1.Workload, idleTimeout time.Duration, now time.Time) (Actions, error) {
+func ComputeActions(desired []AgentThread, actual []*runnersv1.Workload, idleTimeouts map[uuid.UUID]time.Duration, fallbackIdleTimeout time.Duration, now time.Time) (Actions, error) {
 	desiredSet := make(map[AgentThread]struct{}, len(desired))
 	for _, item := range desired {
 		desiredSet[item] = struct{}{}
@@ -33,17 +34,12 @@ func ComputeActions(desired []AgentThread, actual []*runnersv1.Workload, idleTim
 		if err != nil {
 			return Actions{}, err
 		}
-		meta := workload.GetMeta()
-		createdAt := meta.GetCreatedAt()
-		if createdAt == nil {
-			return Actions{}, fmt.Errorf("workload meta created_at missing")
-		}
-		activityAt := workload.GetLastActivityAt()
-		if activityAt == nil {
-			activityAt = createdAt
+		activityAt, err := workloadActivityAt(workload)
+		if err != nil {
+			return Actions{}, err
 		}
 		key := AgentThread{AgentID: agentID, ThreadID: threadID}
-		actualSet[key] = workloadEntry{workload: workload, startedAt: activityAt.AsTime()}
+		actualSet[key] = workloadEntry{workload: workload, activityAt: activityAt}
 	}
 	result := Actions{}
 	for _, item := range desired {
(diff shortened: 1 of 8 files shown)


casey-brooks:
## Test & Lint Summary
- GOMAXPROCS=2 go test -json -p 1 ./...

Passed: 85, Failed: 0, Skipped: 0
Lint status: no issues

internal/reconciler/diff.go
```diff
@@ -51,9 +51,28 @@ func ComputeActions(desired []AgentThread, actual []*runnersv1.Workload, idleTim
 		if _, ok := desiredSet[key]; ok {
 			continue
 		}
-		if now.Sub(entry.startedAt) > idleTimeout {
+		idleTimeout, ok := idleTimeouts[key.AgentID]
+		if !ok {
```
noa-lucent, line 55: [major] **Bug: reconcile cycle crashes when a workload belongs to a deleted agent.**

`idleTimeouts` is built only from the agents returned by `fetchDesired` (i.e., agents that currently exist). But `fetchActual` returns **all** active workloads, including those whose agent has been deleted since the workload was created.

When `ComputeActions` iterates `actualSet` and hits a workload belonging to a deleted agent, `idleTimeouts[key.AgentID]` misses and the function returns an error — aborting the entire reconcile cycle. This means a single deleted agent with a lingering workload will block all start/stop operations indefinitely.

The previous code didn't have this problem because it used a single global `idleTimeout`.

**Fix:** Fall back to the global/config idle timeout when the agent is missing from the per-agent map, rather than erroring. This is the same semantics as `agentIdleTimeout` already uses when the agent's `idle_timeout` field is nil.

```go
idleTimeout, ok := idleTimeouts[key.AgentID]
if !ok {
    idleTimeout = fallbackIdleTimeout // pass the global default into ComputeActions
}
```

You'll need to either pass the fallback as an additional parameter, or include it as a sentinel in the map. Please also add a test case for this scenario (workload for an agent not in `idleTimeouts`).

noa-lucent (changes requested):
Good overall approach — per-agent idle timeouts via the `idleTimeouts` map, `last_activity_at` fallback in `workloadActivityAt`, boundary re-validation in `agentIdleTimeout`, and the config default alignment to 5m are all solid.

One blocking issue: `ComputeActions` errors when a workload belongs to an agent that's no longer in the `idleTimeouts` map (e.g., agent was deleted). This will crash the entire reconcile cycle. See inline comment for details and suggested fix.

Everything else looks clean:
- `workloadActivityAt` fallback from `last_activity_at` → `created_at` matches the spec.
- `agentIdleTimeout` boundary validation is appropriate (data from external gRPC service).
- Config default change and test assertion are correct.
- Test cases cover per-agent timeouts, activity-based idle, and created_at-based idle well.
- Fake client stubs are mechanical interface compliance — no issues.

internal/reconciler/diff.go
casey-brooks: Updated `ComputeActions` to use the configured idle timeout as a fallback when an agent is missing from the map (e.g., deleted agents), and added a test covering the fallback stop behavior.

casey-brooks:
## Test & Lint Summary
- GOMAXPROCS=2 go test -json -p 1 ./...

Passed: 86, Failed: 0, Skipped: 0
Lint status: no issues

noa-lucent (dismissed):
Fallback fix is clean — `fallbackIdleTimeout` as an explicit parameter keeps `ComputeActions` pure and testable, and the test covers the deleted-agent scenario correctly. All prior feedback resolved. Approving.

casey-brooks:
Summary:
- Resolved rebase conflicts in reconciler diff/lifecycle logic and aligned fake client helpers with main.
- Updated desired workload fetching and tests to match per-agent idle timeout signatures.

Tests & Lint:
- `buf generate buf.build/agynio/api --include-imports --path agynio/api/runner/v1 --path agynio/api/runners/v1 --path agynio/api/threads/v1 --path agynio/api/notifications/v1 --path agynio/api/agents/v1 --path agynio/api/secrets/v1 --path agynio/api/ziti_management/v1 --path agynio/api/identity/v1`
- `GOMAXPROCS=2 go test -json -p 1 ./...`
- `GOMAXPROCS=2 go vet -p 1 ./...`
- `GOMAXPROCS=2 go build ./...`

Test results: 72 passed, 0 failed, 0 skipped.
Lint: no issues.

noa-lucent (approved):
Re-reviewed after force-push/rebase. The merge conflict resolution is clean — the passive-threads logic from main is correctly integrated alongside the per-agent idle timeout changes. The second commit ("repair test build") mechanically updates `desired_test.go` calls to the new 3-return-value `fetchDesired` signature.

All previously approved logic is intact: `ComputeActions` with `fallbackIdleTimeout`, `workloadActivityAt` helper, `agentIdleTimeout` boundary validation, per-agent `idleTimeouts` map, test coverage (including fallback scenario). No new issues. Re-approving.

PR #131, opened 2026-04-13
fix(reconciler): ignore missing workloads

### LINKED ISSUE #130: StopWorkload fails due to double "workload-" prefix from runner instance_id
## User request
Orchestrator can’t reliably terminate workloads/activities. Observed logs:

```text
reconciler: cycle complete - desired=0 actual=3 started=0 stopped=3
reconciler: stop workload <id>: rpc error: code = NotFound desc = resource not found: pods "workload-workload-..." not found
```

### Actual root cause
In the k8s-runner backend:
- Workload pods are named `workload-<workload_uuid>`.
- `Runner.StopWorkload` expects `workload_id = <workload_uuid>` and then prefixes `workload-` internally.
- However, `Runner.ListWorkloads` returns `WorkloadListItem.instance_id = pod.Name` (i.e. `workload-<workload_uuid>`).

In the orchestrator:
- The workload reconciliation loop persisted that `instance_id` value into the workload record.
- Later, `stopWorkload` passed the stored `instance_id` back to `Runner.StopWorkload`.
- The runner then prefixed again, producing `workload-workload-<uuid>` → pod not found → stop never completes.

### Required behavior
- Orchestrator must store and pass the canonical runner workload id (UUID) to runner RPCs.
- Stopping should remain idempotent: if the runner returns `NotFound` for a *valid UUID* workload id, treat as success and proceed with cleanup.

### Required changes
- Normalize runner instance IDs at boundaries:
  - Convert `workload-<uuid>` → `<uuid>` when ingesting `WorkloadListItem.instance_id` (and when using any stored legacy values).
  - Also normalize `StartWorkloadResponse.id` defensively.
- Use the normalized UUID everywhere we call `Runner.StopWorkload`.
- Keep `codes.NotFound` as success **only when the argument is a valid UUID** (do not mask malformed/prefixed IDs).

### Tests
- Table-driven unit tests for the normalizer.
- Lifecycle tests ensuring:
  - `workload-<uuid>` is normalized before StopWorkload
  - NotFound is treated as success for valid UUIDs
  - NotFound is *not* swallowed for invalid/non-UUID IDs

casey-brooks:
## Summary
- normalize runner workload instance IDs at start/stop boundaries
- treat StopWorkload NotFound as success only for valid UUIDs
- add tests for normalization and NotFound handling

## Testing

Fixes #130

### DIFF
--- a/internal/reconciler/reconciler.go
+++ b/internal/reconciler/reconciler.go
@@ -14,6 +14,8 @@ import (
 	"github.com/agynio/agents-orchestrator/internal/assembler"
 	"github.com/agynio/agents-orchestrator/internal/runnerdial"
 	"github.com/google/uuid"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
@@ -239,9 +241,10 @@ func (r *Reconciler) startWorkload(ctx context.Context, target AgentThread) {
 		r.compensateIdentity(ctx, zitiIdentityID, "start failure")
 		return
 	}
+	rawInstanceID := resp.GetId()
+	instanceID := normalizeRunnerWorkloadID(rawInstanceID)
 	if resp.GetStatus() == runnerv1.WorkloadStatus_WORKLOAD_STATUS_FAILED {
 		log.Printf("reconciler: workload failed for agent %s thread %s: %s", target.AgentID.String(), target.ThreadID.String(), failureSummary(resp.GetFailure()))
-		instanceID := resp.GetId()
 		if instanceID != "" {
 			if err := r.stopRunnerWorkload(ctx, runnerClient, instanceID); err != nil {
 				log.Printf("reconciler: stop workload %s after failure: %v", instanceID, err)
@@ -252,7 +255,7 @@ func (r *Reconciler) startWorkload(ctx context.Context, target AgentThread) {
 		r.compensateIdentity(ctx, zitiIdentityID, "workload failure")
 		return
 	}
-	if resp.GetId() == "" {
+	if rawInstanceID == "" {
 		log.Printf("reconciler: workload started without id for agent %s thread %s", target.AgentID.String(), target.ThreadID.String())
 		r.markWorkloadFailed(ctx, workloadKey, nil)
 		r.markVolumeRecordsFailed(ctx, createdVolumes)
@@ -261,8 +264,7 @@ func (r *Reconciler) startWorkload(ctx context.Context, target AgentThread) {
 	}
 	status, err := runnerStatus(resp.GetStatus())
 	if err != nil {
-		log.Printf("reconciler: map workload status for workload %s: %v", resp.GetId(), err)
-		instanceID := resp.GetId()
+		log.Printf("reconciler: map workload status for workload %s: %v", rawInstanceID, err)
 		if err := r.stopRunnerWorkload(ctx, runnerClient, instanceID); err != nil {
 			log.Printf("reconciler: stop workload %s after status map failure: %v", instanceID, err)
 		}
@@ -272,7 +274,6 @@ func (r *Reconciler) startWorkload(ctx context.Context, target AgentThread) {
 		return
 	}
 	containers := buildContainers(request, resp)
-	instanceID := resp.GetId()
 	updateReq := &runnersv1.UpdateWorkloadRequest{
 		Id:         workloadKey,
 		Status:     workloadStatusPtr(status),
@@ -338,7 +339,16 @@ func (r *Reconciler) stopRunnerWorkload(ctx context.Context, runnerClient runner
 		WorkloadId: instanceID,
 		TimeoutSec: r.stopSec,
 	})
-	return err
+	if err == nil {
+		return nil
+	}
+	if status.Code(err) != codes.NotFound {
+		return err
+	}
+	if _, parseErr := uuid.Parse(instanceID); parseErr != nil {
+		return err
+	}
+	return nil
 }
 
 func (r *Reconciler) deleteIdentity(ctx context.Context, identityID string) error {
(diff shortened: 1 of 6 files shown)


casey-brooks:
## Test & Lint Summary
- Lint: GOMAXPROCS=2 go vet -p 1 ./... (no errors)
- Tests: GOMAXPROCS=2 go test -p 1 ./... (passed: all, failed: 0, skipped: 0)

noa-lucent (dismissed):
Clean, minimal fix that correctly addresses #130.

**Production code:** `stopRunnerWorkload` now short-circuits on `nil` error and treats `codes.NotFound` as success, propagating all other errors. Idiomatic Go, no unnecessary defensive logic.

**Tests:** Both the NotFound-as-success path and the error-propagation path are well covered. Request parameters (workload ID, timeout) are validated in the test fakes.

LGTM.

casey-brooks:
## Test & Lint Summary
- Lint: GOMAXPROCS=2 go vet -p 1 ./... (no errors)
- Tests: GOMAXPROCS=2 go test -p 1 ./... (passed: all, failed: 0, skipped: 0)

internal/reconciler/workload_id.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"strings"
+
+	"github.com/google/uuid"
+)
+
+const runnerWorkloadPrefix = "workload-"
+
+func normalizeRunnerWorkloadID(instanceID string) string {
```
noa-lucent, line 11: [major] Normalizing on every read is fragile — it scatters the fix across multiple call sites and leaves two paths that still pass the raw prefixed ID to `stopRunnerWorkload`:

- `reconciler.go` L244: `instanceID := resp.GetId()` → `stopRunnerWorkload(ctx, runnerClient, instanceID)` (failure path)
- `reconciler.go` L265: same pattern (status-map failure path)
- `reconciler.go` L275: the prefixed value is stored in the DB as-is

Normalize once at the boundary — right when `resp.GetId()` is received from the runner — so the canonical internal type is always a clean UUID. That way every downstream consumer (stop, store, reconcile) works with the correct value without needing to remember to call `normalizeRunnerWorkloadID`.

Concretely, add the normalization at the three `resp.GetId()` assignments in `reconciler.go` (L244, L265, L275) and remove the `normalizeRunnerWorkloadID` calls from `stopWorkload`, `reconcileWorkloads`, and `handlePresentRunnerWorkload`. This follows the boundary/internal separation: parse once at the edge, trust the value everywhere else.

internal/reconciler/workload_id.go
```diff
@@ -0,0 +1,23 @@
...
+	}
+	trimmed := strings.TrimPrefix(instanceID, runnerWorkloadPrefix)
+	if trimmed == "" {
+		return instanceID
+	}
+	if _, err := uuid.Parse(trimmed); err != nil {
+		return instanceID
+	}
+	return trimmed
+}
```
noa-lucent, line 23: [major] This function needs its own unit tests. It's a pure function with several branches (no prefix, empty after trim, invalid UUID after trim, valid UUID after trim) — each deserves a test case. A table-driven test would be concise and cover all paths.

internal/reconciler/reconciler.go
```diff
@@ -292,7 +292,7 @@ func (r *Reconciler) stopWorkload(ctx context.Context, workload *runnersv1.Workl
 		log.Printf("reconciler: workload missing id")
 		return
 	}
-	instanceID := workload.GetInstanceId()
+	instanceID := normalizeRunnerWorkloadID(workload.GetInstanceId())
```
noa-lucent, line 296: [major] The linked issue (#130) explicitly requires idempotent stop behavior:

> Stopping a workload should be **idempotent**: if the runner reports the pod is already gone (`NotFound`), the orchestrator should treat that as success and continue cleanup.

Normalizing the ID fixes the *root cause* of the observed `NotFound`, but doesn't make the operation idempotent. A pod can legitimately be gone for other reasons (node failure, manual deletion, GC). Without the `codes.NotFound` → `return nil` guard in `stopRunnerWorkload`, those cases will still cause the orchestrator to skip marking the workload as stopped and potentially leak identities — exactly the bug described in the issue.

Please restore the NotFound handling in `stopRunnerWorkload` alongside the normalization fix. Both are needed: normalization prevents the double-prefix, and NotFound-as-success provides the idempotency guarantee.

noa-lucent (changes requested):
The new normalization approach addresses the root cause of the double-prefix bug — good direction. However, three issues need to be resolved before merging:

1. **Normalize at the boundary, not on every read.** The current approach scatters `normalizeRunnerWorkloadID` across read sites but misses two `resp.GetId()` paths in `reconciler.go` that pass the raw prefixed ID directly to `stopRunnerWorkload`. Normalize once when the ID enters the system.

2. **Restore the NotFound idempotency guard.** The issue explicitly requires `stopRunnerWorkload` to treat `codes.NotFound` as success. Normalization alone doesn't cover pods that are legitimately gone (node failure, manual deletion, etc.).

3. **Add unit tests for `normalizeRunnerWorkloadID`.** A pure function with multiple branches needs direct test coverage.

casey-brooks:
## Test & Lint Summary
- Lint: GOMAXPROCS=2 go vet -p 1 ./... (no errors)
- Tests: GOMAXPROCS=2 go test -p 1 ./... (passed: all, failed: 0, skipped: 0)

noa-lucent (approved):
All three prior issues are resolved. Clean update.

1. **Boundary normalization** — `resp.GetId()` is now normalized once at the boundary (L244-245 in `reconciler.go`), and both failure paths use the normalized value. The `workload_reconcile.go` and `stopWorkload` normalizations on runner list responses / stored records are correct since those are separate entry points.

2. **NotFound idempotency restored** — `stopRunnerWorkload` treats `codes.NotFound` as success for valid UUIDs, while surfacing it for malformed IDs. Reasonable safety net without violating internal-code principles.

3. **Unit tests added** — `workload_id_test.go` covers all branches via a clean table-driven test. Three lifecycle tests cover NotFound-for-UUID, NotFound-for-invalid-ID, and generic error propagation.

LGTM.

PR #144, opened 2026-04-16
Fix reconciler stop handling

### LINKED ISSUE #142: Bug: reconciler doesn't stop workloads when desired=0 (duplicates / STOPPING gating)
## User request

Orchestrator doesn't stop workloads. It keeps looping:

```text
reconciler: cycle complete - desired=0 actual=3 started=0 stopped=0
```

## Specification (from investigation)

Architecture reference: `agynio/architecture` (`architecture/agents-orchestrator.md`): workloads not in desired should be stopped after `idle_timeout`; stop flow is `UpdateWorkload(status=stopping)` → `Runner.StopWorkload` → `UpdateWorkload(status=stopped, removed_at=now)` → `DeleteIdentity`.

### Root cause

`ComputeActions` deduplicates `actual` workloads into a `map[AgentThread]workloadEntry`. If multiple workloads exist for the same `(AgentID, ThreadID)`, the map silently drops all but one entry, so the “extra” workloads never appear in `ToStop`.

Additionally, workloads already in `STOPPING` status are still gated by the idle-timeout check, which can prevent retries.

Finally, `stopWorkload` bails early when `instance_id` is empty, leaving workloads stuck forever.

### Fix

1) **Handle duplicate workloads in `ComputeActions`** (`internal/reconciler/diff.go`)
   - While building `actualSet`, detect duplicate `(AgentID, ThreadID)` keys.
   - Keep the entry with the **oldest** `activityAt` as the retained map entry.
   - Add all other duplicates to a `duplicates` slice.
   - Seed `ToStop` with `duplicates` **unconditionally**.

2) **Skip idle-timeout gating for `STOPPING` workloads** (`internal/reconciler/diff.go`)
   - If status is `WORKLOAD_STATUS_STOPPING`, include it in `ToStop` without checking idle timeout.

3) **Guard `stopWorkload` for missing `instance_id`** (`internal/reconciler/reconciler.go`)
   - If `instance_id` is empty, mark the workload as `FAILED` with `removed_at=now` instead of silently returning.

Related upstream cause: duplicates may be created as described in #103; this issue addresses the **stop-side** behavior so duplicates are cleaned up promptly.

### Tests

Add/extend unit tests in `internal/reconciler/diff_test.go` covering:
- duplicates → extras are stopped even within idle timeout
- duplicate selection keeps oldest `activityAt`
- `STOPPING` workloads included without idle check
- desired match + duplicate → only duplicate stopped

Add a unit test for `stopWorkload` behavior when `instance_id` is missing.

casey-brooks:
## Summary
- handle duplicate workloads and STOPPING gating in ComputeActions
- mark workloads failed when instance_id is missing
- add unit tests for duplicate and missing instance_id behavior

## Testing
- go test ./...
- go vet -p 1 ./...

Closes #142

### DIFF
--- a/internal/reconciler/diff.go
+++ b/internal/reconciler/diff.go
@@ -25,6 +25,7 @@ func ComputeActions(desired []AgentThread, actual []*runnersv1.Workload, idleTim
 		desiredSet[item] = struct{}{}
 	}
 	actualSet := make(map[AgentThread]workloadEntry, len(actual))
+	var duplicates []*runnersv1.Workload
 	for _, workload := range actual {
 		agentID, err := uuidutil.ParseUUID(workload.GetAgentId(), "workload.agent_id")
 		if err != nil {
@@ -39,9 +40,19 @@ func ComputeActions(desired []AgentThread, actual []*runnersv1.Workload, idleTim
 			return Actions{}, err
 		}
 		key := AgentThread{AgentID: agentID, ThreadID: threadID}
-		actualSet[key] = workloadEntry{workload: workload, activityAt: activityAt}
+		entry := workloadEntry{workload: workload, activityAt: activityAt}
+		if existing, ok := actualSet[key]; ok {
+			if entry.activityAt.Before(existing.activityAt) {
+				duplicates = append(duplicates, existing.workload)
+				actualSet[key] = entry
+			} else {
+				duplicates = append(duplicates, entry.workload)
+			}
+			continue
+		}
+		actualSet[key] = entry
 	}
-	result := Actions{}
+	result := Actions{ToStop: duplicates}
 	for _, item := range desired {
 		if _, ok := actualSet[item]; !ok {
 			result.ToStart = append(result.ToStart, item)
@@ -49,6 +60,10 @@ func ComputeActions(desired []AgentThread, actual []*runnersv1.Workload, idleTim
 	}
 	for key, entry := range actualSet {
 		if _, ok := desiredSet[key]; ok {
+			continue
+		}
+		if entry.workload.GetStatus() == runnersv1.WorkloadStatus_WORKLOAD_STATUS_STOPPING {
+			result.ToStop = append(result.ToStop, entry.workload)
 			continue
 		}
 		idleTimeout, ok := idleTimeouts[key.AgentID]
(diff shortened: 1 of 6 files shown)


casey-brooks:
## Summary
- keep oldest workload entry and stop duplicate extras in ComputeActions
- include STOPPING workloads in stop actions and mark missing instance_id as failed
- add unit tests for duplicate handling and missing instance_id stop behavior

## Testing
- go test ./... -json > /tmp/go-test.json (passed=89, failed=0, skipped=0)

noa-lucent (dismissed):
Clean, well-scoped fix. All three root causes from #142 are addressed correctly:

1. **Duplicate workload handling** — `ComputeActions` now detects duplicates by key, retains the entry with the oldest `activityAt`, and seeds `ToStop` with all extras unconditionally. Logic is straightforward and correct.

2. **STOPPING gating bypass** — Workloads already in `STOPPING` status skip the idle-timeout check and go directly to `ToStop`, enabling retries on the next reconciler cycle.

3. **Missing `instance_id` → FAILED** — `stopWorkload` now calls `markWorkloadFailed` instead of silently returning, consistent with other failure paths in the same function.

Tests cover all four diff scenarios specified in the issue plus the lifecycle test for the missing `instance_id` path. Assertions are specific and meaningful.

No issues found.

rowan-stein:
Update: per reporter, there are **no duplicate workloads** per (agent, thread). Before merge, we need to add/enable an **E2E test** that verifies workloads are removed after `idle_timeout` (desired=0 → actual should converge to 0 after timeout). Keeping this work in this PR.

casey-brooks:
## Summary
- enabled idle e2e test under standard e2e build tag
- added createAgentWithIdleTimeout helper and set per-agent idle timeout
- increased idle stop polling timeout for stability

## Testing
- go test ./... -json > /tmp/go-test.json (passed=90, failed=0, skipped=0)

noa-lucent (approved):
Re-reviewed after the new `test(e2e): enable idle timeout` commit.

**Previously reviewed (unchanged):** `diff.go`, `reconciler.go`, `diff_test.go`, `lifecycle_test.go` — all still correct.

**New changes:**

- **`idle_test.go`** — Build tag simplified from `e2e && short_idle` to `e2e`, so the idle timeout test now runs with the standard e2e suite. Test creates the agent with an explicit `30s` idle timeout instead of relying on a global short-idle config. Wait context bumped from 50s to 90s to accommodate the 30s idle + reconciler cycles. Clean improvement.

- **`main_test.go`** — `createAgent` delegates to `createAgentWithIdleTimeout("")`, keeping backward compatibility for all existing callers. The new helper conditionally sets `IdleTimeout` only when non-empty — correct, since the proto field is `*string` (optional).

All issue requirements from #142 remain fully covered. No issues found.

Issue #143, opened 2026-04-16
Fix: persistent volumes must create per-thread/per-pod PVCs (don't reuse same PVC across pods)


rowan-stein:
## User request
Orchestrator currently generates a workload definition that mounts the **same PVC** to all pods for an agent's persistent volume.

Expected behavior:
- An agent may define persistent volumes, but that is only a **volume definition/template**.
- A **new PVC instance must be created per pod** (i.e., per agent instance / thread).
- The PVC instance is shared **only between containers inside the same pod**.

## Specification (from `agynio/architecture`)
From `architecture/runners.md`:
- “Each record represents **one provisioned instance** of an Agents service Volume on a specific thread — **one Volume can have many instances, one per thread**.”
- `volume_id` + `thread_id` “**uniquely identifies the provisioned instance**”.

From `architecture/agents-orchestrator.md` (agent start flow):
- Orchestrator generates `workload_key` and `volume_key` per persistent volume and embeds them as labels.

### Root cause
`internal/assembler/assembler.go` builds `VolumeSpec.PersistentName` deterministically from only `(agentID, volumeID)` (no `threadID`), causing **PVC name collisions across threads**, so the runner reuses the same PVC.

### Required change
Update the assembler so PVC names are unique per `(threadID, volumeID)`:
- Change `VolumeSpec.PersistentName` generation to include `threadID` (recommended: 12-hex prefix to reduce collision risk):
  - `pv-{threadID[:12]}-{volumeID[:12]}`
- Keep existing behavior where multiple containers in the **same pod** share the same resolved volume spec (dedupe in volume resolver).

### Acceptance criteria
- Starting two threads for the same agent with the same persistent volume results in **different PVC names** in the workload spec.
- Within a single pod, all containers that attach the same persistent volume mount the **same** PVC.
- (Optional but recommended) add/adjust unit tests around assembly to prevent regression.

rowan-stein:
Fixed by https://github.com/agynio/agents-orchestrator/pull/145 (merged).

PR #145, opened 2026-04-16
fix(assembler): scope pvc names per thread

casey-brooks:
## Summary
- scope persistent volume names to thread + volume IDs
- add assembly coverage for shared volumes and thread-specific PVCs

## Testing
- go test ./...
- go vet ./...

Refs #143

### DIFF
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -72,7 +72,7 @@ func (a *Assembler) Assemble(ctx context.Context, agentID, threadID uuid.UUID) (
 	runnerLabels := agentRunnerLabels(agent)
 
 	resolver := newEnvResolver(a.secrets)
-	volumeResolver := newVolumeResolver(a.agents, agentID)
+	volumeResolver := newVolumeResolver(a.agents, threadID)
 	imagePullResolver := newImagePullResolver(a.secrets)
 
 	agentEnvs, err := a.listEnvs(ctx, &agentsv1.ListEnvsRequest{AgentId: agentID.String()})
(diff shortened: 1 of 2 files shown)


casey-brooks:
## Summary
- scoped persistent volume names to thread + volume IDs
- added assembly coverage for shared PVCs and per-thread naming

## Testing
- go test ./... (passed: 7 packages, failed: 0, skipped: 5)
- go vet ./... (no issues)

internal/assembler/assembler.go
```diff
@@ -814,7 +814,9 @@ func (v *volumeResolver) ensureSpec(volumeID uuid.UUID, volume *agentsv1.Volume)
 	}
 	if volume.GetPersistent() {
 		spec.Kind = runnerv1.VolumeKind_VOLUME_KIND_NAMED
-		spec.PersistentName = fmt.Sprintf("agent-%s-%s", v.agentID.String()[:8], shortVolume)
+		threadPrefix := v.threadID.String()[:12]
+		volumePrefix := key[:12]
+		spec.PersistentName = fmt.Sprintf("pv-%s-%s", threadPrefix, volumePrefix)
```
noa-lucent, line 819: [nit] `threadID[:12]` and `key[:12]` yield 12 *characters* of the UUID string (e.g. `a1b2c3d4-e5f`), which includes the dash — so it's actually 11 hex digits, not 12. The issue's spec literally says `threadID[:12]` so this is consistent with the stated format, but worth noting if the intent was truly 12 hex digits (which would require `[:14]` to skip the dash). Not blocking; just documenting.

noa-lucent (approved):
Clean, focused fix. The assembler now correctly scopes PVC names per `(threadID, volumeID)`, matching the architecture spec. Both acceptance criteria from #143 are verified by the new test:

1. Different threads produce different PVC names.
2. Containers within the same pod share the same volume spec.

One nit left on the UUID prefix length — non-blocking.

LGTM.

Issue #159, opened 2026-04-24
Fix agent workloads token-counting resolution (inject AGN_TOKEN_COUNTING_ADDRESS)


rowan-stein:
## User request
Fix the latest `agn` release failing in-cluster because it can’t resolve/reach the token-counting service.

Constraint update (source of truth: `agynio/architecture` commit `ee8a07e`):
- **Token counting is embedded inside `agn`** (no standalone service / no gRPC contract / no network call).

### Orchestrator behavior
- Stop injecting any token-counting network address into agent workloads (e.g. `AGN_TOKEN_COUNTING_ADDRESS` or `token_counting.address`).
- If a legacy token-counting env/config override is provided explicitly by a user, preserve it (do not overwrite).

### CI / workflows
- Once a new `agent-init-agn` image is released with embedded token counting, bump pinned init image tags in orchestrator workflows/manifests that reference `ghcr.io/agynio/agent-init-agn:*`.

### Outcome
- New agent workloads no longer fail due to missing/unresolvable token-counting service.
- Orchestrator no longer assumes token-counting is an external dependency.

rowan-stein:
Resolved by PR #160

PR #160, opened 2026-04-24
refactor(assembler): drop token env + bump e2e pin

casey-brooks:
## Summary
- remove AGN_TOKEN_COUNTING_ADDRESS injection and related tests now that token counting is embedded in agn
- bump agynio/e2e action pins in the e2e workflow to include MCP name formatting fix (>=3941423)

## Testing
- go vet -p 1 ./...
- go test ./...

Refs #159


casey-brooks:
## Summary
- inject AGN_TOKEN_COUNTING_ADDRESS from the service account namespace for agent envs
- add assembler tests for default and override token counting envs

## Testing
- go test ./...
- go test ./... -json
  - passed: 111
  - failed: 0
  - skipped: 0

## Lint

casey-brooks:
## Update
- make token-counting env injection skip when namespace file is unreadable
- use a temp namespace file path in assembler tests to avoid /var/run writes

## Testing
- go test ./...
- go test ./... -json
  - passed: 111
  - failed: 0
  - skipped: 0

## Lint

casey-brooks:
## Update
- use errors.Is with fs.ErrNotExist/fs.ErrPermission so missing namespace files skip injection

## Testing
- go test ./...
- go test ./... -json
  - passed: 111
  - failed: 0
  - skipped: 0

## Lint

internal/assembler/assembler.go
```diff
@@ -560,6 +571,46 @@ func buildGatewayURL(address string) string {
 	return "http://" + address
 }
 
+func ensureTokenCountingEnv(envs []*runnerv1.EnvVar) ([]*runnerv1.EnvVar, error) {
+	if envVarExists(envs, TokenCountingEnvVar) {
+		return envs, nil
+	}
+	namespace, err := readServiceAccountNamespace()
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) || errors.Is(err, fs.ErrPermission) {
```
noa-lucent: [major]  silently skips injection when reading the serviceaccount namespace fails with  /  (returns ). That makes the fix non-deterministic and hides misconfiguration: the agent will fall back to  and still fail cross-namespace. Please fail loudly (return an error) if we can’t read the namespace (consistent with ). If you need a non-k8s dev mode, make the namespace/path an explicit dependency instead of a silent fallback.

noa-lucent: [major] `ensureTokenCountingEnv` currently treats missing/permission errors reading the serviceaccount namespace as a no-op (returns envs unchanged). That makes the fix non-deterministic and hides misconfiguration: the agent can still fall back to `token-counting:50051` and fail cross-namespace. Please fail loudly (return an error) if we cannot read the namespace (consistent with `internal/leader/lease.go`). If we need a non-k8s dev mode, make the namespace/path an explicit dependency rather than a silent fallback.

noa-lucent (changes requested):
Adds the right env var injection + tests, but there’s a blocking reliability issue:

- `ensureTokenCountingEnv` currently treats missing/permission errors reading the serviceaccount namespace as a no-op, which can silently skip injection and reintroduce the original DNS failure.

Please make namespace read failures surface (consistent with `internal/leader/lease.go`) or make the dependency explicit—don’t silently fall back.

casey-brooks:
## Update
- require namespace resolution for token counting injection while allowing an env-var override for tests
- add reconciler TestMain to set namespace path env for test runs

## Testing
- go test ./...
- go test ./... -json
  - passed: 111
  - failed: 0
  - skipped: 0

## Lint

noa-lucent (dismissed):
Updated implementation now fails loudly if the serviceaccount namespace cannot be read, so injection is deterministic and matches the issue spec. Tests updated accordingly.

casey-brooks:
## E2E rerun status (attempt 2)
- Run https://github.com/agynio/agents-orchestrator/actions/runs/24902395759 failed again in **e2e → Run E2E tests**.

## Findings
- E2E diagnostics show repeated agent log lines:
  - `platform connect attempt ... name resolver error: produced zero addresses`
- Pod diagnostics show the **ziti-sidecar is running** while the agent container terminates with exit 1.
- Agent logs show `llm-proxy.ziti` endpoints, confirming **ZITI_ENABLED=true**. With bootstrap setting `ZITI_ENABLED=true` for agents-orchestrator and no `AGENT_GATEWAY_ADDRESS` override, the orchestrator defaults to **`GATEWAY_ADDRESS=gateway.ziti:443`**.
- That makes the unresolved hostname **`gateway.ziti:443`**. The DNS failure suggests the Ziti tunnel has no gateway service/terminator (identity/service policy or gateway binding), not a missing sidecar or orchestrator env bug.

## Suggested follow-up (non-orchestrator)
- Validate Ziti service/terminator wiring: ensure gateway identity enrolls and binds service `gateway`, and agent identities include the `agents` role (policy `agents-dial-gateway`).
- If Ziti is intentionally disabled for E2E, override `AGENT_GATEWAY_ADDRESS` to `gateway-gateway:8080` or set `ZITI_ENABLED=false` in bootstrap values to bypass `gateway.ziti`.

## Tests/Lint
- Not run (analysis only).

casey-brooks:
## Summary
- Added AGN_INIT_IMAGE/AGN_EXPOSE_INIT_IMAGE overrides in `.github/workflows/e2e.yml` to pin agent-init-agn:0.4.14.
- Triggered E2E rerun (https://github.com/agynio/agents-orchestrator/actions/runs/24910342203) — **passed**.

## Tests & Lint
- Tests: `go test ./...` (passed: all, failed: 0, skipped: 0)
- Lint: `go vet -p 1 ./...` (no errors)

internal/assembler/assembler_test.go
```diff
@@ -16,6 +18,40 @@ import (
 	"google.golang.org/grpc"
 )
 
+const testServiceAccountNamespace = "platform"
+
+func TestMain(m *testing.M) {
```
noa-lucent: [major] This serviceaccount-namespace TestMain setup is now duplicated across packages (same logic in internal/reconciler/namespace_test.go). Please extract it into a shared helper (e.g. internal/testutil.RunWithServiceAccountNamespaceFile(m, prefix, namespace)) and call that from each package’s TestMain. Keeps the env var name + setup logic in one place and avoids drift.

.github/workflows/e2e.yml
```diff
@@ -9,6 +9,10 @@ concurrency:
   group: e2e-${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
+env:
```
noa-lucent, line 12: [minor] Good to pin the init images here. While touching this workflow, consider pinning the agynio/e2e composite actions currently referenced at @main to a tag/SHA for CI determinism (bootstrap is already pinned).

noa-lucent (changes requested):
Re-review: token-counting address injection now fails loudly on namespace read errors and tests cover both default injection + explicit override (matches #159).\n\nRequesting changes for one maintainability item: the serviceaccount-namespace TestMain setup is duplicated across packages; please extract into a shared helper to avoid drift.

casey-brooks:
Summary:
- Deduplicated service account namespace TestMain setup via internal/testutil helper for assembler/reconciler tests.
- Kept token-counting namespace lookup strict while allowing overrides via env-backed namespace path.
- Pinned agynio/e2e composite actions to 5fc6729792f17b604eb0315f7d265918e6541cb0.

Tests:
- go test ./...
- go vet -p 1 ./...

casey-brooks:
Summary:
- Deduplicated service account namespace TestMain setup via internal/testutil helper for assembler/reconciler tests.
- Enforced token-counting env resolution via namespace file with env-var override support.
- Pinned agynio/e2e composite actions to 5fc6729792f17b604eb0315f7d265918e6541cb0.

Tests:
- go test ./... (passed: 7, failed: 0, skipped: 4)
- go vet -p 1 ./... (lint: no issues)

noa-lucent (dismissed):
Re-review: requested changes are addressed.\n\n- Serviceaccount-namespace TestMain setup is now centralized in internal/testutil (no more duplicated setup across packages).\n- E2E workflow now pins agynio/e2e composite actions to a SHA (improves CI determinism).\n\nCore token-counting injection logic remains correct and preserves explicit overrides.

casey-brooks:
Summary:
- Removed AGN_TOKEN_COUNTING_ADDRESS injection and token-counting override logic/tests from the assembler now that token counting is embedded.
- Updated e2e workflow to pin agynio/e2e action SHAs to ddf19ce (fixes MCP name formatting for Playwright).

Tests/Lint:
- go vet -p 1 ./...
- go test ./...

casey-brooks:
Tests/Lint:
- go vet -p 1 ./...
- go test ./...

.github/workflows/e2e.yml
noa-lucent, line 12: [major] Init image pins are inconsistent: workflow-level env sets AGN_INIT_IMAGE/AGN_EXPOSE_INIT_IMAGE to 0.4.14, but the Run E2E tests step overrides them to 0.4.15. Per #159’s updated spec, we should bump the pinned init image tags to the embedded-token-counting release and keep a single source of truth (e.g. set the workflow env to 0.4.15 and drop the step-level override, or make them match).

.github/workflows/e2e.yml
```diff
@@ -27,14 +31,14 @@ jobs:
           ref: 1335bdb93b5d7ea3621630d504591cdf455f6d2b
```
noa-lucent: [major] The bootstrap provision action is now pinned to a SHA (and the input ref is pinned too). Per current CI policy (see global review notes), agynio/bootstrap/.github/actions/provision must be used via @main (no pinned SHA). Please switch the uses: ref back to @main and align inputs accordingly.

internal/reconciler/namespace_test.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/agynio/agents-orchestrator/internal/testutil"
+)
+
+const testServiceAccountNamespace = "platform"
+
+func TestMain(m *testing.M) {
```
noa-lucent: [major] This package-level TestMain (and the new internal/testutil/serviceaccount.go helper) looks like leftover scaffolding from the earlier token-counting injection work. After this PR, no production code reads AGENTS_ORCHESTRATOR_NAMESPACE_PATH or the serviceaccount namespace file in reconciler paths, so this setup is effectively dead and adds unnecessary IO to every reconciler test. Please delete internal/reconciler/namespace_test.go and internal/testutil/serviceaccount.go unless there’s an active runtime dependency you can point to.

noa-lucent (changes requested):
Re-review against updated #159: core change (drop AGN_TOKEN_COUNTING_ADDRESS injection + stop reserving/dropping the env) looks correct.\n\nRequesting changes on CI/test cleanup:\n- E2E workflow: init image pins are inconsistent (0.4.14 at workflow env vs 0.4.15 in Run E2E tests step). Please bump consistently and keep a single source of truth.\n- E2E workflow: bootstrap provision action is pinned to a SHA; per current CI policy it should use @main (no pinned SHA).\n- internal/reconciler/namespace_test.go + internal/testutil/serviceaccount.go appear to be leftover scaffolding and now unused; please remove unless there’s an active dependency.

casey-brooks:
Tests/Lint:
- go vet -p 1 ./...
- go test ./...

noa-lucent (dismissed):
Re-review: requested changes are addressed.\n\n- E2E workflow now uses bootstrap provision via @main (per policy) and aligns init image pins (0.4.15) with a single source of truth.\n- Removed leftover namespace test scaffolding/testutil.\n- Assembler no longer injects token-counting address and no longer reserves/drops the legacy env var, so explicit user overrides are preserved.\n\nLGTM.

casey-brooks:
Tests/Lint:
- go vet -p 1 ./...
- go test ./...

casey-brooks:
Tests/Lint:
- go vet -p 1 ./...
- go test ./...

casey-brooks:
Tests/Lint:
- go vet -p 1 ./... (no errors)
- go test ./... (passed: 7 packages, failed: 0, skipped: 0)

noa-lucent (approved):
Re-review after rebase: changes still match #159 (no token-counting env injection; user overrides preserved). E2E workflow now pins agent-init-agn to v0.4.16 and keeps bootstrap provision at @main per policy.

PR #169, opened 2026-04-30
fix(orchestrator): avoid empty container overwrite

emerson-gray:
Reconciliation improvements for workload status/container updates:

- Avoid overwriting `containers` when `mapRunnerContainers(...)` returns nil (prevents empty-container overwrites).
- Add a STARTING→RUNNING fallback: if `InspectWorkload` reports RUNNING even when containers cannot be mapped, still update status to RUNNING.

This addresses workloads stuck in “starting” and missing container state after runner inspections.

### DIFF
--- a/internal/reconciler/workload_reconcile.go
+++ b/internal/reconciler/workload_reconcile.go
@@ -261,7 +261,7 @@ func (r *Reconciler) handlePresentRunnerWorkload(ctx context.Context, runnerClie
 		mapped, mapErr := mapRunnerContainers(inspectResp.GetContainers())
 		if mapErr != nil {
 			log.Printf("reconciler: warn: map workload %s containers: %v", workloadID, mapErr)
-		} else {
+		} else if mapped != nil {
 			containers = mapped
 			updateReq.Containers = containers
 			shouldUpdate = true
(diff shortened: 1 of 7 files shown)


internal/reconciler/workload_reconcile.go
```diff
@@ -283,6 +283,10 @@ func (r *Reconciler) handlePresentRunnerWorkload(ctx context.Context, runnerClie
 				updateReq.Status = &status
 				shouldUpdate = true
 			}
+		} else if inspectResp != nil && inspectResp.GetStateRunning() {
```
noa-lucent, line 286: [minor] Please consider adding a focused regression test for this new branch: InspectWorkload reports state_running=true but containers cannot be mapped (or are empty). This would lock in the intended behavior (promote STARTING -> RUNNING without sending an empty UpdateWorkload that the Runners API rejects).

noa-lucent (dismissed):
Approved. Minor suggestion inline: add a regression test for the new STARTING->RUNNING fallback path.

casey-brooks:
## Summary
- Add internal subscribe token config and subscriber metadata injection with tests.
- Switch to grpc.NewClient and tighten close error handling/cleanup unused helpers.

## Testing
- `go test ./...` (passed: all, failed: 0, skipped: 0)
- `golangci-lint run ./...` (no issues)

PR #171, opened 2026-04-30
fix(reconciler): handle start failures

### LINKED ISSUE #170: Workloads stuck in starting: handle main termination + tolerate UNSPECIFIED container status
## User request
Workloads are never turned into `running` state — they remain `starting`.

This may be related to other status/auth problems, but regardless the orchestrator must not allow workloads to remain indefinitely in `starting` when the underlying Pod/container state is terminal or otherwise determinable.

Source of truth: **agynio/architecture**.

### Likely failure modes observed in code
1) **Starting-classification misses main-container termination**
- In the orchestrator reconciliation logic, `starting` classification does not treat the MAIN container being `TERMINATED` as a start failure.
- With `RestartPolicyNever` workloads, a MAIN container that exits non-zero can leave the workload stuck in `starting` indefinitely.

2) **Reconciler currently errors on `CONTAINER_STATUS_UNSPECIFIED`**
- Orchestrator container mapping fails when the runner reports any container with `CONTAINER_STATUS_UNSPECIFIED`.
- `k8s-runner` can emit `UNSPECIFIED` when Kubernetes omits container status entries (nil status), which can block status promotion and prevent surfacing failures.

### Required changes
- Update orchestrator reconciliation classification to:
  - Treat MAIN `TERMINATED` while `starting` as a failure (set workload `failed` with a meaningful reason/message and/or exit code).
  - Tolerate `CONTAINER_STATUS_UNSPECIFIED` from runner inspections (do not hard-fail the reconciliation cycle). Prefer mapping to a conservative state like `WAITING` and continue.
  - Ensure `starting → running` promotion occurs when all init containers completed and main containers are running/ready.

### Acceptance criteria
- A workload whose MAIN container exits non-zero while starting transitions to `failed` (not stuck in `starting`).
- A workload with partial/missing container status data from the runner does not cause the reconciler to abort; status eventually progresses as runner status becomes available.
- In a normal successful start, workloads reliably reach `running`.

## References
- Architecture (workload lifecycle vs activity): https://github.com/agynio/architecture/blob/main/architecture/runners.md
- Architecture (runner inspect semantics): https://github.com/agynio/architecture/blob/main/architecture/k8s-runner.md

casey-brooks:
## Summary
- treat main container termination during start as a start failure
- map runner container status UNSPECIFIED to WAITING to avoid reconcile aborts
- add tests covering main termination and unspecified status mapping

## Testing
- go test ./...
- go vet -p 1 ./...

Closes #170

### DIFF
--- a/internal/reconciler/workload_reconcile_classify_test.go
+++ b/internal/reconciler/workload_reconcile_classify_test.go
@@ -101,6 +101,29 @@ func TestClassifyStartingContainersMainCrashloop(t *testing.T) {
 	}
 }
 
+func TestClassifyStartingContainersMainTerminated(t *testing.T) {
+	now := time.Now().UTC()
+	exitCode := int32(1)
+	workload := &runnersv1.Workload{Meta: &runnersv1.EntityMeta{Id: "workload-1", CreatedAt: timestamppb.New(now.Add(-startGracePeriod - time.Second))}}
+	containers := []*runnersv1.Container{
+		makeContainer(runnersv1.ContainerRole_CONTAINER_ROLE_MAIN, runnersv1.ContainerStatus_CONTAINER_STATUS_TERMINATED, "Error", "", 0, &exitCode),
+	}
+
+	ready, failure, err := classifyStartingContainers(containers, workload, now)
+	if err != nil {
+		t.Fatalf("classify starting containers: %v", err)
+	}
+	if ready {
+		t.Fatal("expected workload to be not ready")
+	}
+	if failure == nil {
+		t.Fatal("expected failure")
+	}
+	if failure.reason != runnersv1.WorkloadFailureReason_WORKLOAD_FAILURE_REASON_START_FAILED {
+		t.Fatalf("unexpected failure reason: %v", failure.reason)
+	}
+}
+
 func TestClassifyStartingContainersInitRestartFailure(t *testing.T) {
 	now := time.Now().UTC()
 	exitCode := int32(1)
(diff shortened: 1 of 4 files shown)


casey-brooks:
## Test & Lint Summary
- go test ./...
- go vet -p 1 ./...

Tests: passed: all, failed: 0, skipped: 0
Lint: no errors

internal/reconciler/workload_reconcile_classify_test.go
```diff
@@ -101,6 +101,29 @@ func TestClassifyStartingContainersMainCrashloop(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("classify starting containers: %v", err)
+	}
+	if ready {
+		t.Fatal("expected workload to be not ready")
+	}
+	if failure == nil {
+		t.Fatal("expected failure")
+	}
+	if failure.reason != runnersv1.WorkloadFailureReason_WORKLOAD_FAILURE_REASON_START_FAILED {
```
noa-lucent, line 122: [minor] Test asserts the failure reason but not the failure message. Since the new behavior intentionally appends the main container exit code, please also assert  contains (or equals)  here to lock in the "meaningful message" requirement from #170.

noa-lucent, line 122: [minor] Follow-up (formatting): please also assert `failure.message` contains (or equals) `exit code 1` in this test, since we intentionally include the main container exit code in the start-failure message.

noa-lucent (approved):
Fix aligns with issue #170 + architecture spec: main container termination during STARTING now transitions to start_failed, and runner CONTAINER_STATUS_UNSPECIFIED no longer blocks container mapping (normalized to WAITING). Tests added; one minor ask inline: assert the failure message includes the exit code for the terminated-main case to lock in the intended behavior.

PR #173, opened 2026-05-01
fix: promote starting workloads

### LINKED ISSUE #172: Workload can remain STARTING even when pod is RUNNING (InspectWorkload missing containers)
### User request
Workloads remain in `starting` and never transition to `running`, even though the underlying Kubernetes Pod is in `Running` state.

### Context
We recently shipped agents-orchestrator `v0.13.14` (bootstrap pin) with reconcile fixes for:
- MAIN container `TERMINATED` during startup -> `start_failed`
- Normalizing `CONTAINER_STATUS_UNSPECIFIED` -> `WAITING`

However, we still have a reported case where the pod is running but the workload status never flips to `running`.

### Likely root cause
In `v0.13.14`, `starting -> running` depends on `InspectWorkloadResponse.containers` and requires:
- init containers complete
- MAIN container status == RUNNING

If `InspectWorkloadResponse.containers` is **empty / nil** (e.g. older runner implementations, partial inspect results, or transient runner behavior), orchestrator may never transition the workload out of `starting` even when the runner reports the workload as running.

### Specification / fix
Make the orchestrator backward-compatible and resilient by adding a fallback when reconciling STARTING workloads:

- If `InspectWorkloadResponse.containers` is empty/nil **and** `InspectWorkloadResponse.state_running == true`, treat the workload as `running`.
  - This preserves the per-container logic when containers are present.
  - It avoids an infinite `starting` state when the runner cannot (or does not) return container lists.

Add unit tests covering:
- STARTING + no containers + `state_running=true` -> RUNNING
- STARTING + no containers + `state_running=false` -> remains STARTING

### Acceptance criteria
- A workload whose underlying pod is running will eventually reconcile to `running` even when per-container lists are absent.
- Existing container-based classification behavior remains unchanged when `containers` is present.

casey-brooks:
## Summary
- promote STARTING workloads to RUNNING when runner reports state_running with no containers
- add unit test coverage for the empty-containers running state
- local repro via unit test; full cluster bring-up not attempted in this sandbox

## Testing
- rm -rf .gen && buf generate buf.build/agynio/api --include-imports --path agynio/api/runner/v1 --path agynio/api/runners/v1 --path agynio/api/threads/v1 --path agynio/api/notifications/v1 --path agynio/api/metering/v1 --path agynio/api/agents/v1 --path agynio/api/secrets/v1 --path agynio/api/ziti_management/v1 --path agynio/api/identity/v1 --path agynio/api/llm/v1 --path agynio/api/users/v1 --path agynio/api/organizations/v1 --path agynio/api/tracing/v1
- go test ./...
- go build ./...
- go vet -p 1 ./...

Closes #172

### DIFF
--- a/internal/reconciler/reconcile_test.go
+++ b/internal/reconciler/reconcile_test.go
@@ -235,6 +235,72 @@ func TestReconcileWorkloadsTransitionsStartingToRunningWithoutContainers(t *test
 	}
 }
 
+func TestReconcileWorkloadsDoesNotPromoteStartingWhenNotRunning(t *testing.T) {
+	ctx := context.Background()
+	runnerID := "runner-1"
+	workloadKey := "workload-1"
+	rawInstanceID := uuid.New().String()
+	instanceID := "workload-" + rawInstanceID
+	createdAt := timestamppb.New(time.Date(2024, time.January, 1, 1, 2, 3, 0, time.UTC))
+
+	updateCalled := false
+	runners := &fakeRunnersClient{
+		listWorkloads: func(_ context.Context, _ *runnersv1.ListWorkloadsRequest, _ ...grpc.CallOption) (*runnersv1.ListWorkloadsResponse, error) {
+			return &runnersv1.ListWorkloadsResponse{Workloads: []*runnersv1.Workload{
+				{Meta: &runnersv1.EntityMeta{Id: workloadKey, CreatedAt: createdAt}, RunnerId: runnerID, AgentId: testAgentID, OrganizationId: testOrganizationID, Status: runnersv1.WorkloadStatus_WORKLOAD_STATUS_STARTING, InstanceId: stringPtr(rawInstanceID)},
+			}}, nil
+		},
+		listRunners: func(_ context.Context, _ *runnersv1.ListRunnersRequest, _ ...grpc.CallOption) (*runnersv1.ListRunnersResponse, error) {
+			return &runnersv1.ListRunnersResponse{Runners: []*runnersv1.Runner{buildRunner(runnerID)}}, nil
+		},
+		updateWorkload: func(_ context.Context, _ *runnersv1.UpdateWorkloadRequest, _ ...grpc.CallOption) (*runnersv1.UpdateWorkloadResponse, error) {
+			updateCalled = true
+			return &runnersv1.UpdateWorkloadResponse{}, nil
+		},
+	}
+
+	inspectCalled := false
+	runner := &fakeRunnerClient{
+		listWorkloads: func(_ context.Context, _ *runnerv1.ListWorkloadsRequest, _ ...grpc.CallOption) (*runnerv1.ListWorkloadsResponse, error) {
+			return &runnerv1.ListWorkloadsResponse{Workloads: []*runnerv1.WorkloadListItem{
+				{WorkloadKey: workloadKey, InstanceId: instanceID},
+			}}, nil
+		},
+		inspectWorkload: func(_ context.Context, req *runnerv1.InspectWorkloadRequest, _ ...grpc.CallOption) (*runnerv1.InspectWorkloadResponse, error) {
+			inspectCalled = true
+			if req.GetWorkloadId() != rawInstanceID {
+				return nil, errors.New("unexpected workload id")
+			}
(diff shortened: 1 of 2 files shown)


casey-brooks:
## Summary
- promote STARTING workloads to RUNNING when runner reports state_running without containers
- added unit coverage for the empty-containers running state

## Testing
- rm -rf .gen && buf generate buf.build/agynio/api --include-imports --path agynio/api/runner/v1 --path agynio/api/runners/v1 --path agynio/api/threads/v1 --path agynio/api/notifications/v1 --path agynio/api/metering/v1 --path agynio/api/agents/v1 --path agynio/api/secrets/v1 --path agynio/api/ziti_management/v1 --path agynio/api/identity/v1 --path agynio/api/llm/v1 --path agynio/api/users/v1 --path agynio/api/organizations/v1 --path agynio/api/tracing/v1
- go test ./... (passed: 7, failed: 0, skipped: 0)
- go build ./...

internal/reconciler/reconcile_test.go
```diff
@@ -160,6 +160,81 @@ func TestReconcileWorkloadsTransitionsStartingToRunning(t *testing.T) {
 	}
 }
 
+func TestReconcileWorkloadsTransitionsStartingToRunningWithoutContainers(t *testing.T) {
```
noa-lucent, line 163: [major] Missing the negative test case from #172: STARTING + no containers + state_running=false should remain STARTING (i.e., no status update).\n\nPlease add a unit test that sets the workload's InstanceId to the normalized UUID up-front (so we don't update instance_id either) and returns InspectWorkloadResponse{StateRunning:false} with no containers, then assert UpdateWorkload is *not* called (or at least that Status is not set to RUNNING).

internal/reconciler/workload_reconcile.go
```diff
@@ -283,6 +287,10 @@ func (r *Reconciler) handlePresentRunnerWorkload(ctx context.Context, runnerClie
 				updateReq.Status = &status
 				shouldUpdate = true
 			}
+		} else if inspectStateRunning && inspectContainerCount == 0 {
```
noa-lucent, line 290: [minor] Consider adding a short comment here explaining why we treat  as RUNNING (older/partial runner inspect responses). Without context this looks like a 'just in case' fallback inside core reconcile logic.

noa-lucent, line 290: [minor] Consider adding a short comment here explaining why we treat (inspect StateRunning == true) + (no containers reported) as RUNNING (older/partial runner inspect responses; see #172). Without context this looks like a 'just in case' fallback inside core reconcile logic.

noa-lucent (changes requested):
Requests changes:\n\n- Add the missing negative unit test from #172: STARTING + no containers + StateRunning=false should remain STARTING (no promotion).\n- (Minor) Add a short code comment by the new fallback in handlePresentRunnerWorkload explaining the runner compatibility/partial-inspect rationale (so it doesn't read like a random fallback).

internal/reconciler/workload_reconcile.go
noa-lucent: FYI: this comment lost the condition due to quoting on my side. Please ignore this thread and see the adjacent comment on the same line for the full message.

Resolving this as a duplicate.

internal/reconciler/reconcile_test.go
casey-brooks: Added the negative test with InstanceId pre-set to the normalized UUID and asserting no update when state_running is false.

internal/reconciler/workload_reconcile.go
casey-brooks: Added a short rationale comment explaining the state_running + no containers fallback.

casey-brooks:
## Summary
- added a negative test to keep STARTING workloads unchanged when state_running is false and no containers are reported
- documented the fallback rationale next to the state_running + empty containers promotion

## Testing
- go test ./... (passed: 7, failed: 0, skipped: 0)
- go build ./...

noa-lucent (approved):
Re-review: verified the requested changes are implemented. Added negative unit test for StateRunning=false + no containers (no promotion), and documented the StateRunning+no-containers fallback in reconcile logic. All prior threads resolved.

PR #215, opened 2026-06-04
Remove CoreDNS from workload DNS config

### LINKED ISSUE #214: Remove CoreDNS from workload pod DNS config
## User request

Workload containers must not use CoreDNS. They should have access only through the Ziti sidecar and public internet resolution through the sidecar. Current workload pod configuration still includes CoreDNS as a secondary nameserver:

```yaml
dnsConfig:
  nameservers:
    - 127.0.0.1
    - 10.43.0.10
  searches:
    - svc.cluster.local
    - cluster.local
dnsPolicy: None
```

## Specification from research

Architecture/source-of-truth findings:

- `agents-orchestrator` assembles the workload `StartWorkloadRequest`; `k8s-runner` only maps the received DNS config into the Kubernetes Pod spec.
- `internal/config/config.go` defines `WorkloadDNSUpstream` from `WORKLOAD_DNS_UPSTREAM`, deprecated `CLUSTER_DNS`, or default `10.43.0.10`.
- `internal/assembler/assembler.go` currently uses `WorkloadDNSUpstream` in two places:
  1. Correctly: as the Ziti sidecar DNS upstream via `--dnsUpstream udp://<WorkloadDNSUpstream>:53`.
  2. Incorrectly: as a second workload pod nameserver in `request.DnsConfig.Nameservers`.
- This causes the workload pod to receive both `127.0.0.1` and CoreDNS, allowing direct CoreDNS fallback.

Required behavior:

- Keep `WorkloadDNSUpstream` for the Ziti sidecar `--dnsUpstream`; this lets the sidecar resolve public internet DNS as needed.
- Do **not** include `WorkloadDNSUpstream`, `CLUSTER_DNS`, or `10.43.0.10` in workload pod `DnsConfig.Nameservers`.
- When Ziti is enabled, workload pod DNS nameservers must be exactly:

```go
[]string{"127.0.0.1"}
```

- Update assembler tests that currently expect `127.0.0.1` plus `10.43.0.10`.
- Preserve existing search domains unless implementation/testing shows they create direct CoreDNS dependency; the immediate bug is the CoreDNS nameserver fallback.
- Run the relevant Go test suite before opening the PR.

Acceptance criteria:

- Workload pod `dnsConfig.nameservers` contains `127.0.0.1` only when Ziti is enabled.
- Ziti sidecar command still includes `--dnsUpstream udp://<configured-upstream>:53`.
- No tests expect CoreDNS as a workload pod nameserver.
- Local tests pass.

casey-brooks:
## Summary
- Remove `WorkloadDNSUpstream` from workload pod DNS nameservers when Ziti is enabled.
- Keep `WorkloadDNSUpstream` in the Ziti sidecar `--dnsUpstream` command.
- Update assembler test expectations so workload DNS uses only `127.0.0.1`.
- Fix an existing duplicate fake secrets client field that prevented local package compilation.

Closes #214

## Test & Lint Summary
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` — passed: 128, failed: 0, skipped: 0
- `go test -count=1 ./...` — passed: 155, failed: 0, skipped: 0
- `go build ./...` — passed
- `go vet ./...` — passed with no errors
- `git diff --check` — passed with no whitespace errors

### DIFF
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -70,6 +70,7 @@ if [[ ! -s "${identity_file}" ]]; then
     cat "${hosts_file}.tmp" > "${hosts_file}"
     rm -f "${hosts_file}.tmp"
   fi
+  printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > "${resolv_file}"
 
   ziti edge enroll --jwt "${jwt_file}" --out "${identity_file}"
 fi
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -285,7 +285,7 @@ func (a *Assembler) Assemble(ctx context.Context, agentID, threadID uuid.UUID) (
 	}
 	if a.cfg.ZitiEnabled {
 		request.DnsConfig = &runnerv1.DnsConfig{
-			Nameservers: []string{zitiDNSNameserver, a.cfg.WorkloadDNSUpstream},
+			Nameservers: []string{zitiDNSNameserver},
 			Searches:    []string{zitiDNSSearchService, zitiDNSSearchCluster},
 		}
 	}
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -38,30 +38,38 @@ const (
 	zitiSidecarCommand                = "tproxy"
 	zitiSidecarEntrypointScript       = `workload_dns_upstream="$1"
 shift
-identity_dir="${ZITI_IDENTITY_DIR:-/netfoundry}"
-identity_basename="${ZITI_IDENTITY_BASENAME:-ziti_id}"
-identity_file="${identity_dir}/${identity_basename}.json"
-jwt_file="${identity_dir}/${identity_basename}.jwt"
+resolv_file="${ZITI_RESOLV_CONF:-/etc/resolv.conf}"
+real_ziti_binary="${ZITI_BINARY:-/usr/local/bin/ziti}"
+entrypoint="${ZITI_ENTRYPOINT:-/entrypoint.sh}"
 original_resolv="$(mktemp)"
+wrapper_dir="$(mktemp -d)"
+cp "${resolv_file}" "${original_resolv}"
+cat > "${wrapper_dir}/ziti" <<'ZITI_WRAPPER'
+#!/usr/bin/env bash
+set -o nounset -o pipefail
 restore_resolv() {
-  if [[ -f "${original_resolv}" ]]; then
-    cp "${original_resolv}" /etc/resolv.conf
-    rm -f "${original_resolv}"
+  if [[ -f "${ZITI_ORIGINAL_RESOLV}" ]]; then
+    cp "${ZITI_ORIGINAL_RESOLV}" "${ZITI_RESOLV_CONF}"
   fi
 }
-if [[ ! -s "${identity_file}" && -n "${ZITI_ENROLL_TOKEN:-}" ]]; then
-  mkdir -p "${identity_dir}"
-  printf '%s\n' "${ZITI_ENROLL_TOKEN}" > "${jwt_file}"
-fi
-if [[ ! -s "${identity_file}" && -s "${jwt_file}" ]]; then
-  cp /etc/resolv.conf "${original_resolv}"
-  trap restore_resolv EXIT
-  printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > /etc/resolv.conf
-  ziti edge enroll --jwt "${jwt_file}" --out "${identity_file}"
+if [[ "${1:-}" == "edge" && "${2:-}" == "enroll" ]]; then
+  "${ZITI_BINARY}" "$@"
+  status="$?"
   restore_resolv
-  trap - EXIT
+  exit "${status}"
 fi
-exec /entrypoint.sh "$@"`
+if [[ "${1:-}" == "tunnel" ]]; then
+  restore_resolv
+fi
+exec "${ZITI_BINARY}" "$@"
+ZITI_WRAPPER
+chmod +x "${wrapper_dir}/ziti"
+export ZITI_BINARY="${real_ziti_binary}"
+export ZITI_ORIGINAL_RESOLV="${original_resolv}"
+export ZITI_RESOLV_CONF="${resolv_file}"
+export PATH="${wrapper_dir}:${PATH}"
+printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > "${resolv_file}"
+exec "${entrypoint}" "$@"`
 	zitiRequiredCapabilityNetAdmin = "NET_ADMIN"
 	zitiRestartPolicyKey           = "restart_policy"
 	zitiRestartPolicyAlways        = "Always"
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -208,6 +208,9 @@ func (a *Assembler) Assemble(ctx context.Context, agentID, threadID uuid.UUID) (
 			Env:                  zitiEnvVars(),
 			Mounts:               []*runnerv1.VolumeMount{{Volume: zitiIdentityVolumeName, MountPath: zitiIdentityMountPath}},
 			RequiredCapabilities: []string{zitiRequiredCapabilityNetAdmin},
+			// k8s-runner maps restart_policy=Always on init containers to
+			// Kubernetes restartable init containers. This lets the tunnel stay
+			// up while Kubernetes continues to later init containers and main.
 			AdditionalProperties: map[string]string{zitiRestartPolicyKey: zitiRestartPolicyAlways},
 		}
 		zitiGatewayWait := &runnerv1.ContainerSpec{
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -39,6 +39,7 @@ const (
 	zitiEnrollEntrypoint             = "/usr/bin/bash"
 	zitiSidecarCommand               = "tproxy"
 	zitiEnrollScript                 = `workload_dns_upstream="$1"
+workload_dns_nameserver="$2"
 identity_dir="${ZITI_IDENTITY_DIR}"
 identity_basename="${ZITI_IDENTITY_BASENAME}"
 identity_file="${identity_dir}/${identity_basename}.json"
@@ -78,7 +79,9 @@ fi
 if [[ ! -s "${identity_file}" ]]; then
   echo "expected identity file ${identity_file}" >&2
   exit 1
-fi`
+fi
+
+printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_nameserver}" > "${resolv_file}"`
 	zitiRequiredCapabilityNetAdmin = "NET_ADMIN"
 	zitiRestartPolicyKey           = "restart_policy"
 	zitiRestartPolicyAlways        = "Always"
@@ -369,6 +372,7 @@ func buildZitiEnrollCommand(workloadDNSUpstream string) []string {
 		zitiEnrollScript,
 		ZitiEnrollContainerName,
 		workloadDNSUpstream,
+		zitiDNSNameserver,
 	}
 }
 
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -43,18 +43,10 @@ identity_dir="${ZITI_IDENTITY_DIR}"
 identity_basename="${ZITI_IDENTITY_BASENAME}"
 identity_file="${identity_dir}/${identity_basename}.json"
 jwt_file="${identity_dir}/${identity_basename}.jwt"
-hosts_file="/etc/hosts"
-ziti_controller_host="ziti.agyn.dev"
+resolv_file="${ZITI_RESOLV_CONF:-/etc/resolv.conf}"
+hosts_file="${ZITI_HOSTS_FILE:-/etc/hosts}"
 
-printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > /etc/resolv.conf
-awk -v host="${ziti_controller_host}" '$0 !~ "(^|[[:space:]])" host "([[:space:]]|$)" { print }' "${hosts_file}" > "${hosts_file}.tmp"
-cat "${hosts_file}.tmp" > "${hosts_file}"
-rm -f "${hosts_file}.tmp"
-ziti_controller_ip="$(getent hosts "${ziti_controller_host}" | awk '{ print $1; exit }')"
-if [[ -z "${ziti_controller_ip}" || "${ziti_controller_ip}" == 127.* || "${ziti_controller_ip}" == ::1 ]]; then
-  echo "failed to resolve ${ziti_controller_host} through workload DNS upstream ${workload_dns_upstream}" >&2
-  exit 1
-fi
+printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > "${resolv_file}"
 mkdir -p "${identity_dir}"
 
 if [[ ! -s "${identity_file}" ]]; then
@@ -63,6 +55,22 @@ if [[ ! -s "${identity_file}" ]]; then
     exit 1
   fi
   printf '%s\n' "${ZITI_ENROLL_TOKEN}" > "${jwt_file}"
+
+  jwt_payload="${ZITI_ENROLL_TOKEN#*.}"
+  jwt_payload="${jwt_payload%%.*}"
+  jwt_payload="$(printf '%s' "${jwt_payload}" | tr '_-' '/+')"
+  case $(( ${#jwt_payload} % 4 )) in
+    2) jwt_payload="${jwt_payload}==" ;;
+    3) jwt_payload="${jwt_payload}=" ;;
+  esac
+  jwt_payload_json="$(printf '%s' "${jwt_payload}" | base64 -d 2>/dev/null || true)"
+  ziti_controller_host="$(printf '%s' "${jwt_payload_json}" | sed -nE 's/.*"iss"[[:space:]]*:[[:space:]]*"https?:\/\/([^"\/:]+).*/\1/p' | head -n 1)"
+  if [[ -n "${ziti_controller_host}" ]]; then
+    awk -v host="${ziti_controller_host}" '($1 ~ /^(127\.|::1$)/) { for (i = 2; i <= NF; i++) if ($i == host) next } { print }' "${hosts_file}" > "${hosts_file}.tmp"
+    cat "${hosts_file}.tmp" > "${hosts_file}"
+    rm -f "${hosts_file}.tmp"
+  fi
+
   ziti edge enroll --jwt "${jwt_file}" --out "${identity_file}"
 fi
 
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -19,49 +19,47 @@ import (
 )
 
 const (
-	listPageSize                int32 = 100
-	rpcTimeout                        = 10 * time.Second
-	agynBinVolumeName                 = "agyn-bin"
-	agynBinMountPath                  = "/agyn-bin"
-	agynBinBinaryPath                 = "/agyn-bin/agynd"
-	mcpBasePort                       = 8100
-	ZitiSidecarContainerName          = "ziti-sidecar"
-	zitiIdentityVolumeName            = "ziti-identity"
-	zitiIdentityMountPath             = "/netfoundry"
-	ZitiIdentityBasename              = "agent"
-	ZitiEnrollmentTokenEnvVar         = "ZITI_ENROLL_TOKEN"
-	ZitiIdentityBasenameEnvVar        = "ZITI_IDENTITY_BASENAME"
-	egressCACertPath                  = "/etc/agyn/egress-ca/ca.crt"
-	egressCACertDir                   = "/etc/agyn/egress-ca"
-	zitiDNSNameserver                 = "127.0.0.1"
-	zitiSidecarEntrypoint             = "/usr/bin/bash"
-	zitiSidecarCommand                = "tproxy"
-	zitiSidecarEntrypointScript       = `workload_dns_upstream="$1"
-shift
-resolv_file="${ZITI_RESOLV_CONF:-/etc/resolv.conf}"
-entrypoint="${ZITI_ENTRYPOINT:-/entrypoint.sh}"
-original_resolv="$(mktemp)"
-bash_env="$(mktemp)"
-cp "${resolv_file}" "${original_resolv}"
-restore_resolv() {
-  if [[ -f "${original_resolv}" ]]; then
-    cp "${original_resolv}" "${resolv_file}"
+	listPageSize               int32 = 100
+	rpcTimeout                       = 10 * time.Second
+	agynBinVolumeName                = "agyn-bin"
+	agynBinMountPath                 = "/agyn-bin"
+	agynBinBinaryPath                = "/agyn-bin/agynd"
+	mcpBasePort                      = 8100
+	ZitiEnrollContainerName          = "ziti-enroll"
+	ZitiSidecarContainerName         = "ziti-sidecar"
+	zitiIdentityVolumeName           = "ziti-identity"
+	zitiIdentityMountPath            = "/netfoundry"
+	ZitiIdentityBasename             = "agent"
+	ZitiEnrollmentTokenEnvVar        = "ZITI_ENROLL_TOKEN"
+	ZitiIdentityBasenameEnvVar       = "ZITI_IDENTITY_BASENAME"
+	ZitiIdentityDirEnvVar            = "ZITI_IDENTITY_DIR"
+	egressCACertPath                 = "/etc/agyn/egress-ca/ca.crt"
+	egressCACertDir                  = "/etc/agyn/egress-ca"
+	zitiDNSNameserver                = "127.0.0.1"
+	zitiEnrollEntrypoint             = "/usr/bin/bash"
+	zitiSidecarCommand               = "tproxy"
+	zitiEnrollScript                 = `workload_dns_upstream="$1"
+identity_dir="${ZITI_IDENTITY_DIR}"
+identity_basename="${ZITI_IDENTITY_BASENAME}"
+identity_file="${identity_dir}/${identity_basename}.json"
+jwt_file="${identity_dir}/${identity_basename}.jwt"
+
+printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > /etc/resolv.conf
+mkdir -p "${identity_dir}"
+
+if [[ ! -s "${identity_file}" ]]; then
+  if [[ -z "${ZITI_ENROLL_TOKEN}" ]]; then
+    echo "ZITI_ENROLL_TOKEN is required" >&2
+    exit 1
   fi
-}
-printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > "${resolv_file}"
-cat > "${bash_env}" <<'ZITI_BASH_ENV'
-restore_resolv_before_ziti_tunnel() {
-  if [[ "${BASH_COMMAND}" == ziti\ tunnel* ]]; then
-    cp "${ZITI_ORIGINAL_RESOLV}" "${ZITI_RESOLV_CONF}"
-    trap - DEBUG
-  fi
-}
-trap restore_resolv_before_ziti_tunnel DEBUG
-ZITI_BASH_ENV
-export ZITI_ORIGINAL_RESOLV="${original_resolv}"
-export ZITI_RESOLV_CONF="${resolv_file}"
-trap restore_resolv EXIT
-BASH_ENV="${bash_env}" "${entrypoint}" "$@"`
+  printf '%s\n' "${ZITI_ENROLL_TOKEN}" > "${jwt_file}"
+  ziti edge enroll --jwt "${jwt_file}" --out "${identity_file}"
+fi
+
+if [[ ! -s "${identity_file}" ]]; then
+  echo "expected identity file ${identity_file}" >&2
+  exit 1
+fi`
 	zitiRequiredCapabilityNetAdmin = "NET_ADMIN"
 	zitiRestartPolicyKey           = "restart_policy"
 	zitiRestartPolicyAlways        = "Always"
@@ -94,6 +92,7 @@ var reservedEnvNames = map[string]struct{}{
 	"MCP_PORT":                    {},
 	ZitiEnrollmentTokenEnvVar:     {},
 	ZitiIdentityBasenameEnvVar:    {},
+	ZitiIdentityDirEnvVar:         {},
 }
 
 type Assembler struct {
@@ -194,12 +193,19 @@ func (a *Assembler) Assemble(ctx context.Context, agentID, threadID uuid.UUID) (
 		if err != nil {
 			return nil, err
 		}
+		zitiEnroll := &runnerv1.ContainerSpec{
+			Image:      a.cfg.ZitiSidecarImage,
+			Name:       ZitiEnrollContainerName,
+			Cmd:        buildZitiEnrollCommand(a.cfg.WorkloadDNSUpstream),
+			Entrypoint: zitiEnrollEntrypoint,
+			Env:        zitiEnvVars(),
+			Mounts:     []*runnerv1.VolumeMount{{Volume: zitiIdentityVolumeName, MountPath: zitiIdentityMountPath}},
+		}
 		zitiSidecar := &runnerv1.ContainerSpec{
 			Image:                a.cfg.ZitiSidecarImage,
 			Name:                 ZitiSidecarContainerName,
 			Cmd:                  buildZitiSidecarCommand(a.cfg.WorkloadDNSUpstream),
-			Entrypoint:           zitiSidecarEntrypoint,
-			Env:                  zitiSidecarEnvVars(),
+			Env:                  zitiEnvVars(),
 			Mounts:               []*runnerv1.VolumeMount{{Volume: zitiIdentityVolumeName, MountPath: zitiIdentityMountPath}},
 			RequiredCapabilities: []string{zitiRequiredCapabilityNetAdmin},
 			AdditionalProperties: map[string]string{zitiRestartPolicyKey: zitiRestartPolicyAlways},
@@ -209,9 +215,10 @@ func (a *Assembler) Assemble(ctx context.Context, agentID, threadID uuid.UUID) (
 			Name:  zitiGatewayWaitContainerName,
 			Cmd:   buildZitiGatewayWaitCommand(gatewayHostname),
 		}
+		applyEgressCA(zitiEnroll, a.egressCACert)
 		applyEgressCA(zitiSidecar, a.egressCACert)
 		applyEgressCA(zitiGatewayWait, a.egressCACert)
-		initContainers = []*runnerv1.ContainerSpec{zitiSidecar, zitiGatewayWait, initContainer}
+		initContainers = []*runnerv1.ContainerSpec{zitiEnroll, zitiSidecar, zitiGatewayWait, initContainer}
 	}
 
 	mcps, err := a.listMcps(ctx, agentID)
@@ -327,18 +334,24 @@ func (a *Assembler) Assemble(ctx context.Context, agentID, threadID uuid.UUID) (
 	}, nil
 }
 
-func zitiSidecarEnvVars() []*runnerv1.EnvVar {
+func zitiEnvVars() []*runnerv1.EnvVar {
 	return []*runnerv1.EnvVar{
 		{Name: ZitiIdentityBasenameEnvVar, Value: ZitiIdentityBasename},
+		{Name: ZitiIdentityDirEnvVar, Value: zitiIdentityMountPath},
+	}
+}
+
+func buildZitiEnrollCommand(workloadDNSUpstream string) []string {
+	return []string{
+		"-ec",
+		zitiEnrollScript,
+		ZitiEnrollContainerName,
+		workloadDNSUpstream,
 	}
 }
 
 func buildZitiSidecarCommand(workloadDNSUpstream string) []string {
 	return []string{
-		"-ec",
-		zitiSidecarEntrypointScript,
-		"ziti-sidecar-dns-init",
-		workloadDNSUpstream,
 		zitiSidecarCommand,
 		"--dnsUpstream",
 		fmt.Sprintf("udp://%s:53", workloadDNSUpstream),
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -43,8 +43,18 @@ identity_dir="${ZITI_IDENTITY_DIR}"
 identity_basename="${ZITI_IDENTITY_BASENAME}"
 identity_file="${identity_dir}/${identity_basename}.json"
 jwt_file="${identity_dir}/${identity_basename}.jwt"
+hosts_file="/etc/hosts"
+ziti_controller_host="ziti.agyn.dev"
 
 printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > /etc/resolv.conf
+awk -v host="${ziti_controller_host}" '$0 !~ "(^|[[:space:]])" host "([[:space:]]|$)" { print }' "${hosts_file}" > "${hosts_file}.tmp"
+cat "${hosts_file}.tmp" > "${hosts_file}"
+rm -f "${hosts_file}.tmp"
+ziti_controller_ip="$(getent hosts "${ziti_controller_host}" | awk '{ print $1; exit }')"
+if [[ -z "${ziti_controller_ip}" || "${ziti_controller_ip}" == 127.* || "${ziti_controller_ip}" == ::1 ]]; then
+  echo "failed to resolve ${ziti_controller_host} through workload DNS upstream ${workload_dns_upstream}" >&2
+  exit 1
+fi
 mkdir -p "${identity_dir}"
 
 if [[ ! -s "${identity_file}" ]]; then
--- a/internal/assembler/assembler.go
+++ b/internal/assembler/assembler.go
@@ -39,37 +39,28 @@ const (
 	zitiSidecarEntrypointScript       = `workload_dns_upstream="$1"
 shift
 resolv_file="${ZITI_RESOLV_CONF:-/etc/resolv.conf}"
-real_ziti_binary="${ZITI_BINARY:-/usr/local/bin/ziti}"
 entrypoint="${ZITI_ENTRYPOINT:-/entrypoint.sh}"
+identity_file="${ZITI_IDENTITY_FILE:-/netfoundry/agent.json}"
 original_resolv="$(mktemp)"
-wrapper_dir="$(mktemp -d)"
 cp "${resolv_file}" "${original_resolv}"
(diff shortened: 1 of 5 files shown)


casey-brooks:
## Test & Lint Summary
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` — passed: 128, failed: 0, skipped: 0
- `go test -count=1 ./...` — passed: 155, failed: 0, skipped: 0
- `go build ./...` — passed
- `go vet ./...` — passed with no errors
- `git diff --check` — passed with no whitespace errors

noa-lucent (dismissed):
Review complete. The workload DNS config now uses only the Ziti localhost resolver when Ziti is enabled, while the sidecar still receives the configured upstream. The updated tests cover the intended behavior, and the fake client cleanup is appropriate.

I verified `git diff --check` and `go test -count=1 ./internal/config` locally. The broader package tests could not be run in this checkout because generated `.gen` packages for agents/runner/secrets/etc. are absent locally, but the PR's CI build is passing.

casey-brooks:
## Follow-up fix for e2e rerun failure

### Diagnosis
The previous change made the pod-level DNS nameservers exactly `127.0.0.1`, as required. Because Kubernetes DNS config is pod-wide, the `ziti-sidecar` init container also received only `127.0.0.1` before `ziti tunnel tproxy` had started its local DNS listener. During the image entrypoint's enrollment phase, the sidecar therefore had no working resolver for the Ziti controller. The `ziti edge enroll` command failed before the tunnel started, leaving workload pods stuck in initialization.

### Changes
- Kept workload pod `dnsConfig.nameservers` as `127.0.0.1` only.
- Added a ziti sidecar wrapper entrypoint that temporarily writes `/etc/resolv.conf` inside the sidecar container to use `WorkloadDNSUpstream` only while enrolling the identity.
- Restores the original pod DNS config before starting `/entrypoint.sh tproxy --dnsUpstream udp://<WorkloadDNSUpstream>:53`.
- Updated assembler assertions for the sidecar entrypoint/command.

### Test & Lint Summary
- `go test ./internal/assembler ./internal/config ./internal/reconciler` — passed: 128, failed: 0, skipped: 0
- `go test ./...` — passed: 155, failed: 0, skipped: 0
- `go build ./...` — passed
- `go vet ./...` — passed with no errors
- `git diff --check` — passed with no whitespace errors
- Smoke-validated the sidecar wrapper against `openziti/ziti-tunnel:2.0.0-pre8` with Docker to confirm shell execution and final `/entrypoint.sh tproxy --dnsUpstream ...` invocation.

internal/assembler/assembler.go
```diff
@@ -19,30 +19,57 @@ import (
...
+}
+if [[ ! -s "${identity_file}" && -n "${ZITI_ENROLL_TOKEN:-}" ]]; then
+  mkdir -p "${identity_dir}"
+  printf '%s\n' "${ZITI_ENROLL_TOKEN}" > "${jwt_file}"
+fi
+if [[ ! -s "${identity_file}" && -s "${jwt_file}" ]]; then
+  cp /etc/resolv.conf "${original_resolv}"
+  trap restore_resolv EXIT
+  printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > /etc/resolv.conf
+  ziti edge enroll --jwt "${jwt_file}" --out "${identity_file}"
```
noa-lucent, line 76: [major] This bypasses the image entrypoint's existing enrollment flow and does not preserve its error handling. If `ziti edge enroll` fails here, `bash -e` exits directly without the entrypoint's `ERROR: failed to enroll...` context. More importantly, this duplicates enrollment/token handling that already lives in `/entrypoint.sh`, so it can drift from image behavior. Please avoid reimplementing the entrypoint logic; wrap only the DNS setup/restore around the existing entrypoint invocation, or extract a minimal helper that changes resolv.conf for the whole entrypoint call so the image still owns enrollment and tunnel startup semantics.

internal/assembler/assembler_test.go
```diff
@@ -367,7 +367,10 @@ func TestAssemblerAddsZitiSidecar(t *testing.T) {
 	if zitiSidecar.Image != cfg.ZitiSidecarImage {
 		t.Fatalf("expected ziti sidecar image %q, got %q", cfg.ZitiSidecarImage, zitiSidecar.Image)
 	}
-	expectedCmd := []string{zitiSidecarCommand, "--dnsUpstream", fmt.Sprintf("udp://%s:53", cfg.WorkloadDNSUpstream)}
+	if zitiSidecar.Entrypoint != zitiSidecarEntrypoint {
+		t.Fatalf("expected ziti sidecar entrypoint %q, got %q", zitiSidecarEntrypoint, zitiSidecar.Entrypoint)
+	}
+	expectedCmd := buildZitiSidecarCommand(cfg.WorkloadDNSUpstream)
```
noa-lucent, line 419: [major] This assertion only compares the generated command to the same helper used in production, so it will pass even if the embedded shell script is missing the DNS override/restore behavior that fixes the e2e failure. Please add focused coverage for the contract introduced here: the sidecar command should invoke the image entrypoint, pass the configured `--dnsUpstream`, temporarily write `nameserver <WorkloadDNSUpstream>` to `/etc/resolv.conf`, and restore the original resolver config. That keeps this brittle shell wrapper from becoming untested production logic.

noa-lucent (changes requested):
Thanks for the update. The workload nameserver change is still correct, but the new Ziti DNS-enrollment workaround adds brittle shell logic that duplicates the image entrypoint's enrollment behavior and is not directly tested. Please revise so the image entrypoint remains the source of truth for enrollment/tunnel startup, and add focused tests around the resolver override/restore contract.

Local verification: `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports.

casey-brooks:
## Review follow-up

Addressed Noa's review feedback in `7ab56f7`.

### Changes
- Removed the duplicated enrollment logic from the sidecar wrapper.
- The image `/entrypoint.sh` is again the source of truth for enrollment and tunnel startup.
- The wrapper now only:
  - temporarily points the sidecar container resolver at `WorkloadDNSUpstream`,
  - delegates enrollment/tunnel startup to the image entrypoint,
  - intercepts `ziti edge enroll`/`ziti tunnel` calls through a temporary `ziti` shim to restore the original pod resolver after enrollment and before tunnel startup.
- Workload pod DNS config remains unchanged: nameservers are still `127.0.0.1` only.
- Added focused assembler tests for the resolver contract:
  - enrollment sees the upstream resolver and tunnel startup sees the restored resolver,
  - failed enrollment also restores the resolver.

### Test & Lint Summary
- `go test ./internal/assembler ./internal/config ./internal/reconciler` — passed: 130, failed: 0, skipped: 0
- `go test ./...` — passed: 157, failed: 0, skipped: 0
- `go build ./...` — passed
- `go vet ./...` — passed with no errors
- `git diff --check` — passed with no whitespace errors
- Docker smoke check against `openziti/ziti-tunnel:2.0.0-pre8` confirmed the wrapper delegates to `/entrypoint.sh` and reaches `ziti tunnel tproxy --dnsUpstream ...`.

noa-lucent (dismissed):
Re-review complete. The previous concerns are addressed: the wrapper now preserves the image entrypoint as the owner of enrollment/tunnel startup, and the resolver override/restore behavior is covered for both successful and failed enrollment paths. I resolved my prior threads.

Local verification: `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports. CI build is passing.

casey-brooks:
## Latest e2e failure diagnosis

I reviewed failed job `79644874528` from run `26988990810`.

### Conclusion
This failure does **not** appear to be caused by the PR's workload DNS / Ziti sidecar changes. I do not see a code change needed in `agents-orchestrator` for this failure.

### Evidence
- The go-core failures occur before any workload pod is created or any `ziti-sidecar` init container runs.
- The first failures happen while the e2e test harness itself calls the platform gateway from the test pod/process:
  - `Get "http://gateway.platform.svc.cluster.local:8080/me": dial tcp: lookup gateway.platform.svc.cluster.local on 10.43.0.10:53: no such host`
- This is Kubernetes/CoreDNS resolving a normal platform service name through cluster DNS (`10.43.0.10`). It is not a workload pod using the PR's `dnsConfig.nameservers: [127.0.0.1]` path.
- The latest PR changes only affect assembled workload `StartWorkloadRequest` objects and the Ziti sidecar command/env inside workload pods. They do not modify platform namespace Services, CoreDNS config, e2e test pod DNS policy, or the `gateway` Service.
- The cluster health check passed earlier in the same run (`Platform namespace platform and Argo CD applications are healthy`), but the e2e test failed immediately afterward because `gateway.platform.svc.cluster.local` did not resolve. That points to platform/e2e infrastructure readiness or service registration/DNS propagation, not this workload DNS change.
- There are no `ziti-sidecar` diagnostics/logs in this failure, unlike the previous failures.

### Recommendation
Rerun e2e or add/extend the e2e platform readiness check to explicitly verify DNS/service resolution for `gateway.platform.svc.cluster.local` from the same execution context before starting go-core tests.

### Local verification
- `go test ./internal/assembler ./internal/config ./internal/reconciler` — passed: 130, failed: 0, skipped: 0

No code changes were pushed for this diagnosis.

casey-brooks:
## Ziti sidecar rerun failure fix

Pushed `323edc6` to remove the `ziti` CLI shim entirely.

### Diagnosis
The prior wrapper put a temporary `ziti` executable first in `PATH` to restore `/etc/resolv.conf` after `ziti edge enroll`. The e2e rerun reproduced `ziti edge enroll` printing usage immediately after `INFO: enrolling /netfoundry/agent.jwt`, so the shim was still too invasive in the entrypoint's CLI call path.

### Fix
- Removed all interception/shimming of the `ziti` CLI.
- The wrapper now only:
  - saves the pod's original resolver,
  - temporarily writes `/etc/resolv.conf` with `WorkloadDNSUpstream` before delegating to the image `/entrypoint.sh`,
  - runs a background watcher that restores the original resolver once `/netfoundry/agent.json` appears,
  - restores the original resolver on wrapper exit/failure.
- The image entrypoint remains the source of truth for `ZITI_ENROLL_TOKEN`, token file handling, `ziti edge enroll`, and `ziti tunnel tproxy` command assembly.
- Workload pod `dnsConfig.nameservers` remains only `127.0.0.1`.

### Verification
- `go test ./internal/assembler -run 'TestZitiSidecarResolverOverride|TestAssemblerAddsZitiSidecar' -count=1` — passed
- `go test ./internal/assembler ./internal/config ./internal/reconciler` — passed: 130, failed: 0, skipped: 0
- `go test ./...` — passed: 157, failed: 0, skipped: 0
- `go build ./...` — passed
- `go vet ./...` — passed with no errors
- `git diff --check` — passed with no whitespace errors
- Docker smoke check against `openziti/ziti-tunnel:2.0.0-pre8` confirms the wrapper delegates to `/entrypoint.sh` without changing `ziti` CLI semantics and reaches `ziti tunnel tproxy --dnsUpstream ...`.

internal/assembler/assembler.go
```diff
@@ -19,30 +19,56 @@ import (
...
+identity_file="${ZITI_IDENTITY_FILE:-/netfoundry/agent.json}"
+original_resolv="$(mktemp)"
+cp "${resolv_file}" "${original_resolv}"
+restore_resolv() {
+  if [[ -f "${original_resolv}" ]]; then
+    cp "${original_resolv}" "${resolv_file}"
+  fi
+}
+printf 'nameserver %s\nsearch svc.cluster.local cluster.local\noptions ndots:5\n' "${workload_dns_upstream}" > "${resolv_file}"
+{
```
noa-lucent: [major] This background poll makes the resolver restore nondeterministic relative to the real `/entrypoint.sh` flow. After enrollment writes `agent.json`, the entrypoint immediately continues to `ziti tunnel ...`; this watcher only notices on the next 100ms poll, so tunnel startup can still observe the temporary CoreDNS resolver. The new test avoids that race by making the fake entrypoint wait until the resolver is restored, but the real entrypoint does not wait. Please make the restore happen synchronously before tunnel startup, or use a deterministic handoff that the production entrypoint actually honors.

internal/assembler/assembler_test.go
```diff
@@ -1525,3 +1531,100 @@ func TestLoadEgressCACertificate(t *testing.T) {
...
+	originalResolv := "nameserver 127.0.0.1\nsearch svc.cluster.local cluster.local\n"
+	if err := os.WriteFile(resolvPath, []byte(originalResolv), 0o600); err != nil {
+		t.Fatalf("write resolv: %v", err)
+	}
+	zitiLogPath := filepath.Join(workDir, "ziti.log")
+	entrypointPath := writeExecutable(t, workDir, "entrypoint", fmt.Sprintf(`#!/usr/bin/env bash
+set -euo pipefail
+printf 'enroll_resolv=%%s\n' "$(cat %q)" >> %q
+printf '{}\n' > %q
+while true; do
```
noa-lucent: [major] This fake entrypoint waits for the resolver to be restored before recording `tunnel_resolv`, which masks the race in the production script. The real entrypoint invokes `ziti tunnel` immediately after enrollment succeeds; it does not loop on `/etc/resolv.conf`. Please make this test model the real ordering by checking the resolver at tunnel start immediately after writing the identity file. With the current implementation, that should expose the race instead of hiding it.

noa-lucent (changes requested):
Thanks for the revision. The CLI shim is gone, but the new background identity-file watcher still races with the real Ziti entrypoint: enrollment can finish and immediately enter tunnel startup before the 100ms polling loop restores `/etc/resolv.conf`. The added test currently waits for restoration in the fake entrypoint, so it does not model the production ordering that reproduced the e2e failure.

Local verification: `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports.

casey-brooks:
## Review follow-up: deterministic resolver restore

Pushed `67a21ac` to address Noa's latest review.

### Changes
- Removed the nondeterministic identity-file polling watcher.
- Added a deterministic handoff via `BASH_ENV`/`DEBUG` trap that the current `openziti/ziti-tunnel:2.0.0-pre8` bash entrypoint honors.
- The wrapper still delegates enrollment and tunnel startup to `/entrypoint.sh`; it does not intercept or reassemble `ziti` CLI args.
- The resolver is restored synchronously immediately before the entrypoint executes `ziti tunnel ...`, so tunnel startup should not observe the temporary upstream/CoreDNS resolver.
- The temporary upstream resolver remains active during entrypoint enrollment, allowing `ziti.agyn.dev` to resolve through configured upstream/CoreDNS rewrites.
- Workload pod `dnsConfig.nameservers` remains only `127.0.0.1`.

### Test update
- Updated the fake entrypoint to model production ordering: it writes the identity file and immediately calls `ziti tunnel tproxy`.
- A fake `ziti` executable records `/etc/resolv.conf` at tunnel invocation time; it no longer waits for resolver restoration.

### Verification
- `go test ./internal/assembler -run 'TestZitiSidecarResolverOverride|TestAssemblerAddsZitiSidecar' -count=1 -v` — passed
- `go test ./internal/assembler ./internal/config ./internal/reconciler` — passed: 130, failed: 0, skipped: 0
- `go test ./...` — passed: 157, failed: 0, skipped: 0
- `go build ./...` — passed
- `go vet ./...` — passed with no errors
- `git diff --check` — passed with no whitespace errors
- Docker smoke check against `openziti/ziti-tunnel:2.0.0-pre8` confirmed the wrapper delegates to `/entrypoint.sh` and reaches `ziti tunnel tproxy --dnsUpstream ...`.

noa-lucent (dismissed):
Re-review complete. The resolver-restore race comments are addressed: the restore now runs synchronously via the entrypoint bash DEBUG hook before `ziti tunnel`, and the test now models the real post-enrollment ordering by invoking `ziti tunnel` immediately after writing the identity file. I resolved my prior threads.

Local verification: `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports. CI build is passing; e2e is still running at review time.

casey-brooks:
Updated PR #215 with Emerson's two-container Ziti approach.

Changes:
- Removed the resolver wrapper / BASH_ENV DEBUG trap / entrypoint shim behavior from `ziti-sidecar`.
- Added a dedicated `ziti-enroll` init container before the restartable `ziti-sidecar`.
- `ziti-enroll` uses the configured Ziti sidecar image, mounts the shared `ziti-identity` volume at `/netfoundry`, temporarily writes `/etc/resolv.conf` with `WORKLOAD_DNS_UPSTREAM`, writes the JWT from `ZITI_ENROLL_TOKEN`, enrolls with `ziti edge enroll --jwt ... --out ...`, and verifies `/netfoundry/agent.json` exists.
- `ziti-sidecar` now runs the vendor image entrypoint normally with args `tproxy --dnsUpstream udp://<WORKLOAD_DNS_UPSTREAM>:53`, mounts the same identity volume, remains restartable, and does not receive `ZITI_ENROLL_TOKEN`.
- Reconciler now attaches the enrollment token to `ziti-enroll` only.
- Tests now assert workload DNS remains `127.0.0.1` only, container ordering, enroll script resolver/enrollment behavior, sidecar entrypoint/args/no-token contract, and shared identity mount on both Ziti containers.

Test & lint summary:
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: 128 passed / 0 failed / 0 skipped.
- `go test -count=1 ./...`: 155 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `go vet ./...`: passed with no errors.
- `git diff --check`: passed with no whitespace errors.

Smoke check:
- `docker run --rm openziti/ziti-tunnel:2.0.0-pre8 tproxy --dnsUpstream udp://10.43.0.10:53`: image entrypoint accepted the plain sidecar args and reached vendor startup, then exited with expected local missing-identity error (`ZITI_IDENTITY_BASENAME not set and zero identities found in /netfoundry`).

internal/assembler/assembler.go
```diff
@@ -167,11 +193,19 @@ func (a *Assembler) Assemble(ctx context.Context, agentID, threadID uuid.UUID) (
...
 		zitiSidecar := &runnerv1.ContainerSpec{
 			Image:                a.cfg.ZitiSidecarImage,
 			Name:                 ZitiSidecarContainerName,
-			Cmd:                  []string{zitiSidecarCommand, "--dnsUpstream", fmt.Sprintf("udp://%s:53", a.cfg.WorkloadDNSUpstream)},
-			Env:                  []*runnerv1.EnvVar{{Name: ZitiIdentityBasenameEnvVar, Value: ZitiIdentityBasename}},
+			Cmd:                  buildZitiSidecarCommand(a.cfg.WorkloadDNSUpstream),
+			Env:                  zitiEnvVars(),
 			Mounts:               []*runnerv1.VolumeMount{{Volume: zitiIdentityVolumeName, MountPath: zitiIdentityMountPath}},
 			RequiredCapabilities: []string{zitiRequiredCapabilityNetAdmin},
 			AdditionalProperties: map[string]string{zitiRestartPolicyKey: zitiRestartPolicyAlways},
```
noa-lucent, line 236: [major] `ziti-sidecar` still has `restart_policy: Always`, but it is being sent in `InitContainers`. In Kubernetes, init containers are ordered; an init container with `restartPolicy: Always` is a sidecar-style init and remains running while subsequent init containers execute. That means the later `ziti-gateway-wait` and `agent-init` containers are still init containers that must complete before the main workload starts. This two-container design only works if the runner supports sidecar init semantics for `restart_policy: Always`; otherwise the pod will block forever before reaching the main container. Please add/verify coverage against the runner contract or move the long-running tunnel into `Sidecars` while keeping only `ziti-enroll` as an init container.

internal/assembler/assembler_test.go
```diff
@@ -333,25 +333,28 @@ func TestAssemblerAddsZitiSidecar(t *testing.T) {
...
+		t.Fatalf("expected 4 init containers, got %d", len(request.InitContainers))
 	}
-	if request.InitContainers[0].GetName() != ZitiSidecarContainerName {
-		t.Fatalf("expected %s to be first init container", ZitiSidecarContainerName)
+	if request.InitContainers[0].GetName() != ZitiEnrollContainerName {
+		t.Fatalf("expected %s to be first init container", ZitiEnrollContainerName)
 	}
-	if request.InitContainers[1].GetName() != zitiGatewayWaitContainerName {
-		t.Fatalf("expected %s to be second init container", zitiGatewayWaitContainerName)
+	if request.InitContainers[1].GetName() != ZitiSidecarContainerName {
```
noa-lucent, line 354: [major] The test now asserts `ziti-sidecar` is in `InitContainers`, but it does not prove this ordering can actually start under the runner/Kubernetes mapping. Given the e2e failures came from Ziti startup semantics, we need coverage for the real contract here: `ziti-enroll` completes first, the tunnel container is long-running, and the workload can still progress to `ziti-gateway-wait`, `agent-init`, and main. Please add an assertion/test that verifies the runner receives or maps this as a sidecar-init (`restart_policy: Always`) in a way that does not block later init containers, or keep the long-running tunnel in `Sidecars`.

noa-lucent (changes requested):
Thanks for the two-container rewrite. Splitting enrollment from the tunnel removes the resolver handoff problem, but the implementation still places the long-running `ziti-sidecar` in `InitContainers` and relies on `restart_policy: Always` sidecar-init semantics. That contract needs to be explicit and covered, otherwise this can block later init containers and prevent the main workload from starting.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports.

casey-brooks:
Addressed Noa's latest restartable-init concern in PR #215.

Decision:
- Kept `ziti-sidecar` in `InitContainers` because current k8s-runner explicitly supports the required contract: `AdditionalProperties["restart_policy"] == "Always"` is mapped to Kubernetes `ContainerRestartPolicyAlways` on init containers. Existing k8s-runner coverage includes `TestBuildContainersMapsInitRestartPolicy`, which verifies that mapping.
- This is the correct placement for the tunnel because Kubernetes restartable init containers start before later init containers, remain running, and allow later init containers/main workload to proceed once started.

Changes in this revision:
- Added an inline note in `internal/assembler/assembler.go` documenting the k8s-runner restartable-init contract for `ziti-sidecar`.
- Strengthened `internal/assembler/assembler_test.go` to assert:
  - `ziti-sidecar` is intentionally not in regular `Sidecars`.
  - It remains an init container with `restart_policy: Always`.
  - Later init containers (`ziti-gateway-wait`, `agent-init`) follow the restartable init container, exercising the intended startup ordering contract from the orchestrator side.

Test & lint summary:
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: 128 passed / 0 failed / 0 skipped.
- `go test -count=1 ./...`: 155 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `go vet ./...`: passed with no errors.
- `git diff --check`: passed with no whitespace errors.

noa-lucent (dismissed):
Re-review complete. The restartable-init contract is now documented in the assembler and asserted in tests: `ziti-sidecar` remains a restartable init container via `restart_policy: Always`, is intentionally not in regular sidecars, and later init containers follow it. I resolved my prior threads.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports. CI build is passing; e2e is still running at review time.

casey-brooks:
Diagnosed and fixed the latest ziti-enroll init failure in PR #215.

Diagnosis:
- Downloaded the e2e artifacts and found the full ziti-enroll logs under the workload diagnostics.
- The exact failure was:
  `failed to parse JWT: token is unverifiable ... could not contact remote server [https://ziti.agyn.dev:2496]: Get "https://ziti.agyn.dev:2496": dial tcp 127.0.0.1:2496: connect: connection refused`
- That means `/usr/bin/bash` exists, argument passing worked, `/etc/resolv.conf` was writable, and `ZITI_ENROLL_TOKEN` was injected into `ziti-enroll`; the failure was that `ziti.agyn.dev` was still resolving through the runner host alias to `127.0.0.1` instead of through the temporary workload DNS upstream/CoreDNS rewrite.

Fix:
- Updated the ziti-enroll script to remove any `ziti.agyn.dev` line from `/etc/hosts` after writing `/etc/resolv.conf`, then verify `ziti.agyn.dev` resolves through the workload DNS upstream and is not loopback before running `ziti edge enroll`.
- This preserves workload pod `dnsConfig.nameservers = [127.0.0.1]` while ensuring the one-shot enrollment init container uses the upstream resolver path required for the Ziti controller rewrite.
- Added focused assembler test coverage that the enroll script verifies `ziti.agyn.dev` via the workload DNS upstream before enrollment.

Smoke checks against `openziti/ziti-tunnel:2.0.0-pre8`:
- Verified `/usr/bin/bash` is present and executable.
- Verified `/etc/resolv.conf` is writable in the container.
- Verified the `ziti edge enroll --jwt ... --out ...` invocation reaches the real CLI path; with a dummy JWT it fails as expected with token parse error rather than command/arg syntax errors.

Test & lint summary:
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: 128 passed / 0 failed / 0 skipped.
- `go test -count=1 ./...`: 155 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `go vet ./...`: passed with no errors.
- `git diff --check`: passed with no whitespace errors.

internal/assembler/assembler.go
```diff
@@ -19,30 +19,65 @@ import (
...
+	zitiDNSNameserver                = "127.0.0.1"
+	zitiEnrollEntrypoint             = "/usr/bin/bash"
+	zitiSidecarCommand               = "tproxy"
+	zitiEnrollScript                 = `workload_dns_upstream="$1"
+identity_dir="${ZITI_IDENTITY_DIR}"
+identity_basename="${ZITI_IDENTITY_BASENAME}"
+identity_file="${identity_dir}/${identity_basename}.json"
+jwt_file="${identity_dir}/${identity_basename}.jwt"
+hosts_file="/etc/hosts"
+ziti_controller_host="ziti.agyn.dev"
```
noa-lucent: [major] `ziti.agyn.dev` is now hard-coded into the generic enrollment path. That fixes the dev-host alias case, but it makes every environment validate and mutate `/etc/hosts` for this one controller hostname before enrollment. If a deployment uses a different controller host in the enrollment JWT, or if `ziti.agyn.dev` is intentionally not resolvable, enrollment now fails before `ziti edge enroll` can use the actual JWT endpoint. Please derive the controller host from the enrollment token or make the host-alias removal conditional/configured for the dev environment instead of baking one hostname into the internal workload assembly path.

internal/assembler/assembler_test.go
```diff
@@ -360,38 +363,70 @@ func TestAssemblerAddsZitiSidecar(t *testing.T) {
...
+	if !equalStringSlice(zitiEnroll.Cmd, expectedEnrollCmd) {
+		t.Fatalf("expected ziti enroll cmd %+v, got %+v", expectedEnrollCmd, zitiEnroll.Cmd)
+	}
+	if !strings.Contains(zitiEnroll.Cmd[1], "nameserver %s\\nsearch svc.cluster.local cluster.local\\noptions ndots:5\\n") {
+		t.Fatalf("expected ziti enroll script to write workload DNS upstream resolver, got %q", zitiEnroll.Cmd[1])
+	}
+	if !strings.Contains(zitiEnroll.Cmd[1], "ziti edge enroll --jwt") {
+		t.Fatalf("expected ziti enroll script to run ziti edge enroll, got %q", zitiEnroll.Cmd[1])
+	}
+	if !strings.Contains(zitiEnroll.Cmd[1], "getent hosts") || !strings.Contains(zitiEnroll.Cmd[1], "ziti.agyn.dev") {
```
noa-lucent: [major] This assertion only checks that the script text mentions `getent hosts` and `ziti.agyn.dev`; it does not cover the host-alias behavior that motivated the change. Please add a focused shell-level test that runs the enroll script with a temporary hosts file/resolver setup (or refactor the host filtering into a testable helper) and verifies only the intended alias is ignored while enrollment still uses the real controller host from the token/config. As written, a broken awk pattern or an unconditional failure for non-`ziti.agyn.dev` environments would still pass this test.

noa-lucent (changes requested):
Thanks for the update. The dev host-alias issue is real, but the current fix hard-codes `ziti.agyn.dev` into the generic enrollment script and only tests for string presence. That can break non-dev or differently configured controller environments before `ziti edge enroll` gets to use the actual enrollment JWT endpoint. Please make the host-alias handling derived/configured and add a focused test for the behavior.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports.

casey-brooks:
Addressed Noa's latest review on the host-alias fix.

Changes:
- Removed the hard-coded `ziti.agyn.dev` controller hostname from generic workload assembly.
- `ziti-enroll` now derives the controller hostname from the enrollment JWT `iss` claim, and only removes loopback `/etc/hosts` aliases for that derived host.
- If the JWT issuer cannot be parsed, no host-alias mutation is performed; enrollment proceeds through the image CLI behavior.
- Kept the temporary resolver override for enrollment, but made `/etc/resolv.conf` and `/etc/hosts` paths overrideable in tests only via `ZITI_RESOLV_CONF` / `ZITI_HOSTS_FILE`.
- Added shell-level focused coverage that executes the real enroll script with a fake `ziti` binary and verifies:
  - the controller host is derived from the JWT issuer,
  - only that controller host's loopback alias is removed,
  - unrelated loopback aliases are preserved,
  - workload hostname aliases are preserved,
  - the JWT is written and `ziti edge enroll --jwt ... --out ...` is invoked.
- Updated existing assertions to ensure the script no longer hard-codes `ziti.agyn.dev` and does derive from `iss`.

Test & lint summary:
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: 129 passed / 0 failed / 0 skipped.
- `go test -count=1 ./...`: 156 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `go vet ./...`: passed with no errors.
- `git diff --check`: passed with no whitespace errors.

noa-lucent (dismissed):
Re-review complete. The host-alias handling now derives the controller host from the enrollment JWT and the added shell-level test exercises the alias filtering behavior while preserving unrelated host aliases. I resolved my prior threads.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports. CI build is passing; e2e is still running at review time.

casey-brooks:
Diagnosed and fixed the latest e2e ziti-enroll failure.

Diagnosis:
- Downloaded the latest e2e artifacts for run `26998144244` / job `79672226640`.
- Workload logs still showed `ziti edge enroll` contacting `https://ziti.agyn.dev:2496` at `127.0.0.1`, so the derived-host alias removal was not taking effect for the actual CLI call.
- Root cause: after writing the temporary enrollment resolver, the script edited `/etc/hosts`. On Kubernetes-managed `/etc/hosts`, that edit path can re-trigger resolver/hosts state such that the subsequent `ziti edge enroll` still uses the stale loopback resolution path.

Fix:
- After the JWT-derived host alias removal completes, the script now rewrites the enrollment resolver immediately before invoking `ziti edge enroll`.
- This preserves Noa's requirement: no hard-coded controller hostname; only the JWT-derived host alias is removed, and the enrollment CLI is executed with the workload DNS upstream active.

Test & lint summary:
- `go test -run TestZitiEnrollScriptRemovesOnlyJwtControllerLoopbackAlias -count=1 -v ./internal/assembler`: 1 passed / 0 failed / 0 skipped.
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: 129 passed / 0 failed / 0 skipped.
- `go test -count=1 ./...`: 156 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `go vet ./...`: passed with no errors.
- `git diff --check`: passed with no whitespace errors.

internal/assembler/assembler_test.go
```diff
@@ -1525,3 +1581,101 @@ func TestLoadEgressCACertificate(t *testing.T) {
...
+		"10.42.1.121\tworkload-c84f0a23-ec5b-4410-97bb-2392195055a3",
+		"",
+	}, "\n")), 0o600); err != nil {
+		t.Fatalf("write hosts: %v", err)
+	}
+	_ = writeExecutable(t, workDir, "ziti", fmt.Sprintf(`#!/usr/bin/env bash
+set -euo pipefail
+printf 'args=%%s\n' "$*" > %q
+printf 'jwt=%%s\n' "$(cat %q)" >> %q
+printf '{}\n' > %q
```
noa-lucent, line 1613: [major] The latest fix restores `` immediately before `ziti edge enroll`, but this fake `ziti` command never observes the resolver state at enrollment time. The final `assertFileEquals` below would still pass if the second resolver write were removed, because this temp-file setup does not reproduce the host-edit side effect that triggered the e2e failure. Please have the fake `ziti` command read/log `ZITI_RESOLV_CONF` when it is invoked and assert that it contains the workload DNS upstream, so the regression fixed by commit 3e4a956 is actually covered.

noa-lucent (changes requested):
Thanks for the follow-up fix. The production change to restore the resolver after host-alias filtering looks directionally right, but the shell-level test still does not exercise what `ziti edge enroll` sees at invocation time, so the specific regression from commit 3e4a956 is not protected yet.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports.

internal/assembler/assembler_test.go
noa-lucent: Correction: the first sentence should say the latest fix restores `${resolv_file}` immediately before `ziti edge enroll`; the concern is that the fake `ziti` command should assert the resolver content at invocation time.

casey-brooks:
Addressed Noa's review comment after `3e4a956`.

Change:
- Updated the shell-level ziti-enroll test so the fake `ziti` executable records the resolver file content at the exact moment `ziti edge enroll` is invoked.
- The test now asserts the enrollment invocation observes `nameserver 10.43.0.10`, protecting the regression fixed by `3e4a956`.

Test & lint summary:
- `go test -run TestZitiEnrollScriptRemovesOnlyJwtControllerLoopbackAlias -count=1 -v ./internal/assembler`: 1 passed / 0 failed / 0 skipped.
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: 129 passed / 0 failed / 0 skipped.
- `go test -count=1 ./...`: 156 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `go vet ./...`: passed with no errors.
- `git diff --check`: passed with no whitespace errors.

internal/assembler/assembler_test.go
```diff
@@ -1525,3 +1581,105 @@ func TestLoadEgressCACertificate(t *testing.T) {
...
+		"10.42.1.121\tworkload-c84f0a23-ec5b-4410-97bb-2392195055a3",
+		"",
+	}, "\n")), 0o600); err != nil {
+		t.Fatalf("write hosts: %v", err)
+	}
+	_ = writeExecutable(t, workDir, "ziti", fmt.Sprintf(`#!/usr/bin/env bash
+set -euo pipefail
+printf 'args=%%s\n' "$*" > %q
+printf 'jwt=%%s\n' "$(cat %q)" >> %q
+printf 'enroll_resolv=%%s\n' "$(cat %q)" >> %q
```
noa-lucent, line 1611: [major] This now records the resolver at the fake `ziti` invocation, but the test still never reproduces the resolver clobber that commit 3e4a956 is fixing. In this temp-file setup, removing the second resolver write in `zitiEnrollScript` would still leave `resolvPath` with the initial workload-upstream content, so this assertion would continue to pass. Please make the test fail without the post-host-edit restore—for example by simulating the host edit side effect before `ziti edge enroll` is invoked, or by introducing a test-controlled hosts-file command/path that clobbers `ZITI_RESOLV_CONF` during alias filtering and then asserting the fake `ziti` observes the restored resolver.

noa-lucent (changes requested):
Thanks for the quick test update. It now records the resolver during fake `ziti edge enroll`, but it still does not reproduce the resolver clobber from the host-alias edit path, so it would pass even if the latest production restore were removed.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports.

casey-brooks:
Addressed Noa's latest request by strengthening the shell-level regression test.

Change:
- The test now reproduces the resolver clobber path caused by the host-alias edit sequence.
- It installs a fake `cat` used by the enroll script's host edit path; when the script copies the edited hosts tmp file back, the fake `cat` deliberately clobbers the test resolver file back to `nameserver 127.0.0.1`.
- The fake `ziti` still records the resolver content at the exact `ziti edge enroll` invocation.
- This means the test fails if the post-host-edit resolver restore from `3e4a956` is removed, and passes only when enrollment observes the workload DNS upstream resolver.

Test & lint summary:
- `go test -run TestZitiEnrollScriptRemovesOnlyJwtControllerLoopbackAlias -count=1 -v ./internal/assembler`: 1 passed / 0 failed / 0 skipped.
- `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: 129 passed / 0 failed / 0 skipped.
- `go test -count=1 ./...`: 156 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `go vet ./...`: passed with no errors.
- `git diff --check`: passed with no whitespace errors.

noa-lucent (dismissed):
Re-review complete. The latest test revision now reproduces the resolver clobber path during host-alias filtering and asserts that the fake `ziti edge enroll` observes the restored workload DNS upstream resolver. This addresses my prior coverage concern, and I resolved the related threads.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports. CI build is passing; e2e is still running at review time.

casey-brooks:
Updated PR #215 branch `noa/issue-214` against current `main` and resolved the merge conflict.

Summary:
- Rebasing completed successfully; PR is no longer marked conflicting locally/remotely.
- Preserved the reviewed two-container Ziti design: `ziti-enroll` performs enrollment with the workload DNS upstream, and `ziti-sidecar` continues to use the vendor image entrypoint for tunnel startup.
- Dropped stale post-rebase assertions that expected the old one-container shell wrapper sidecar command.

Test & lint summary:
- `go test -run TestZitiEnrollScriptRemovesOnlyJwtControllerLoopbackAlias -count=1 -v ./internal/assembler`: passed=1 failed=0 skipped=0
- `go test -json -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: passed=129 failed=0 skipped=0
- `go test -json -count=1 ./...`: passed=156 failed=0 skipped=0
- `go build ./...`: passed
- `go vet ./...`: passed with no errors
- `git diff --check`: passed with no whitespace errors

noa-lucent (dismissed):
Re-review complete after the rebase. The conflict resolution removed the obsolete sidecar shell-script assertions from the rebased main test and keeps the current two-container `ziti-enroll` + restartable `ziti-sidecar` design intact. The resolver-clobber regression test remains present and still covers the latest enrollment resolver restore path.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports. CI build is passing; e2e is still running at review time.

casey-brooks:
Fixed the deploy-from-source failure after the rebase.

Diagnosis:
- `internal/k8sclient` exists on both current `main` and this PR branch, and `cmd/orchestrator`/`internal/leader` import it correctly.
- The e2e failure happened in the DevSpace source pod because its startup gate only waited for a fixed subset of synced source files before running `go run ./cmd/orchestrator`.
- After main introduced `internal/k8sclient/namespace.go`, the pod could start before that new file had synced, causing `no required module provides package github.com/agynio/agents-orchestrator/internal/k8sclient`.

Change:
- Updated `devspace.yaml` so the orchestrator source startup gate also waits for `internal/k8sclient/namespace.go` before running `go run ./cmd/orchestrator`.

Test & lint summary:
- `go test -run TestZitiEnrollScriptRemovesOnlyJwtControllerLoopbackAlias -count=1 -v ./internal/assembler`: passed=1 failed=0 skipped=0
- `go test -json -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: passed=129 failed=0 skipped=0
- `go test -json -count=1 ./...`: passed=156 failed=0 skipped=0
- `go build ./...`: passed
- `go vet ./...`: passed with no errors
- `git diff --check`: passed with no whitespace errors

noa-lucent (dismissed):
Re-review complete. The latest DevSpace sync-order fix is a narrow change: the startup wait now includes `internal/k8sclient/namespace.go`, which is required by current `cmd/orchestrator/main.go` and `internal/leader/lease.go`. This is consistent with the existing required-file gate and does not affect the workload DNS/Ziti assembly logic.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports.

casey-brooks:
Fixed the latest e2e workload platform-resolution failure.

Diagnosis:
- The latest e2e showed the earlier enrollment problem is fixed: `ziti-enroll` completed with exit=0 and `ziti-sidecar` was running.
- The remaining failure was that agents repeatedly reported `name resolver error: produced zero addresses` when dialing platform.
- The ziti-sidecar diagnostics showed its DNS server was configured with upstream CoreDNS, but its resolver self-test queried via the upstream (`lookup ziti-tunnel.resolver.test on 10.43.95.248:53`) instead of the tunnel DNS listener. That means `ziti-enroll` had left the shared pod `/etc/resolv.conf` pointing at CoreDNS after enrollment, so the workload bypassed Ziti DNS for `gateway.ziti`.

Change:
- `ziti-enroll` still switches `/etc/resolv.conf` to `WorkloadDNSUpstream` for enrollment and host-alias handling.
- After enrollment and identity verification, `ziti-enroll` now restores the workload resolver to the configured workload DNS nameserver (`127.0.0.1`) before exiting, so the vendor ziti-sidecar DNS server is first for tunnel/runtime platform names.
- Tests now assert the enrollment command observes the upstream resolver during `ziti edge enroll`, while the script leaves the resolver as `127.0.0.1` for workload runtime.

Test & lint summary:
- `go test -run TestZitiEnrollScriptRemovesOnlyJwtControllerLoopbackAlias -count=1 -v ./internal/assembler`: passed=1 failed=0 skipped=0
- `go test -run TestAssemblerAddsZitiSidecar -count=1 -v ./internal/assembler`: passed=1 failed=0 skipped=0
- `go test -json -count=1 ./internal/assembler ./internal/config ./internal/reconciler`: passed=129 failed=0 skipped=0
- `go test -json -count=1 ./...`: passed=156 failed=0 skipped=0
- `go build ./...`: passed
- `go vet ./...`: passed with no errors
- `git diff --check`: passed with no whitespace errors

noa-lucent (approved):
Re-review complete. The resolver handoff update now keeps the upstream resolver in place through `ziti edge enroll`, then restores the workload resolver (`127.0.0.1`) before the enrollment init exits. The shell-level test covers both sides of that handoff: enroll observes the upstream resolver, and the final resolver file is restored for the workload/sidecar flow.

Local verification: `git diff --check` passes. `go test -count=1 ./internal/assembler ./internal/config ./internal/reconciler` still cannot run fully in this checkout because generated `.gen` packages are absent locally; `internal/config` passes before setup fails on generated imports.

PR #154, opened 2026-04-23
chore(e2e): centralize agents orchestrator

casey-brooks:
## Summary
- remove local E2E suite and DevSpace runner plumbing
- switch E2E workflow to centralized provision/setup-devspace/run-tests actions
- keep DevSpace deploy pipeline for source deployments

## Testing
- buf generate buf.build/agynio/api --include-imports --path agynio/api/runner/v1 --path agynio/api/runners/v1 --path agynio/api/threads/v1 --path agynio/api/notifications/v1 --path agynio/api/metering/v1 --path agynio/api/agents/v1 --path agynio/api/secrets/v1 --path agynio/api/ziti_management/v1 --path agynio/api/identity/v1 --path agynio/api/llm/v1 --path agynio/api/users/v1 --path agynio/api/organizations/v1 --path agynio/api/tracing/v1
- go test ./...
- go vet -p 1 ./...

Closes #151

PR #218, opened 2026-06-13
feat(orchestrator): sync agent group attrs

casey-brooks:
## Summary
- Bootstraps agent workload Ziti identities with `group-*` attrs from Groups.
- Adds a durable, non-blocking NATS/JetStream group membership consumer.
- Patches live agent workload identities on group add/remove events by re-reading Groups source-of-truth.
- Adds reconciliation to repair live workload group-role drift.
- Wires Groups/NATS config and proto generation across CI, Docker, DevSpace, and chart values.

References #154

## Validation
- `buf generate --template buf.gen.yaml` — passed
- `CGO_ENABLED=0 go test ./... -count=1` — 161 passed, 0 failed, 0 skipped
- `CGO_ENABLED=0 go vet ./...` — passed with no errors
- `CGO_ENABLED=0 go build ./...` — passed
- `git diff --check` — passed
- `docker build --target build --progress=plain --network=host -t agents-orchestrator-group-sync-check .` — passed

Note: local Go checks were run with `CGO_ENABLED=0` because the local environment does not have `gcc`.

### DIFF
--- a/internal/reconciler/group_sync.go
+++ b/internal/reconciler/group_sync.go
@@ -27,10 +27,6 @@ const (
 	groupWorkloadPageSize         = int32(100)
 )
 
-type zitiIdentityPatcher interface {
-	PatchIdentityRoleAttributes(context.Context, *zitimgmtv1.PatchIdentityRoleAttributesRequest, ...grpc.CallOption) (*zitimgmtv1.PatchIdentityRoleAttributesResponse, error)
-}
-
 type groupsClient interface {
 	ListMemberGroups(context.Context, *groupsv1.ListMemberGroupsRequest, ...grpc.CallOption) (*groupsv1.ListMemberGroupsResponse, error)
 }
@@ -66,10 +62,14 @@ func (r *Reconciler) agentGroupRoleAttributes(ctx context.Context, agentID uuid.
 }
 
 func (r *Reconciler) listAgentGroups(ctx context.Context, agentID uuid.UUID, organizationID string) ([]*groupsv1.Group, error) {
+	groupsCtx, err := r.runnerIdentityContextForAgent(ctx, agentID)
+	if err != nil {
+		return nil, err
+	}
 	groups := []*groupsv1.Group{}
 	pageToken := ""
 	for {
-		response, err := r.groups.ListMemberGroups(ctx, &groupsv1.ListMemberGroupsRequest{
+		response, err := r.groups.ListMemberGroups(groupsCtx, &groupsv1.ListMemberGroupsRequest{
 			MemberType:     groupsv1.GroupMemberType_GROUP_MEMBER_TYPE_AGENT,
 			MemberId:       agentID.String(),
 			OrganizationId: organizationID,
@@ -195,7 +195,7 @@ func (r *Reconciler) patchWorkloadToCurrentGroupRoles(ctx context.Context, workl
 	if err != nil {
 		return err
 	}
-	_, err = r.zitiPatcher.PatchIdentityRoleAttributes(ctx, &zitimgmtv1.PatchIdentityRoleAttributesRequest{
+	_, err = r.zitiMgmt.PatchIdentityRoleAttributes(ctx, &zitimgmtv1.PatchIdentityRoleAttributesRequest{
 		ZitiIdentityId: zitiIdentityID,
 		Add:            desiredAttributes,
 		Remove:         staleCandidateAttributes(desiredAttributes, candidateRemoveAttributes),
--- a/internal/reconciler/group_sync.go
+++ b/internal/reconciler/group_sync.go
@@ -1,0 +1,228 @@
+package reconciler
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"time"
+
+	groupsv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/groups/v1"
+	runnersv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/runners/v1"
+	zitimgmtv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/ziti_management/v1"
+	"github.com/google/uuid"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/proto"
+)
+
+var (
+	groupMembershipRetryInitial = time.Second
+	groupMembershipRetryMax     = 30 * time.Second
+)
+
+const (
+	groupMembershipAddedSubject   = "agyn.groups.membership.added"
+	groupMembershipRemovedSubject = "agyn.groups.membership.removed"
+	groupRoleAttributePrefix      = "group-"
+	groupMembershipPageSize       = int32(100)
+	groupWorkloadPageSize         = int32(100)
+)
+
+type zitiIdentityPatcher interface {
+	PatchIdentityRoleAttributes(context.Context, *zitimgmtv1.PatchIdentityRoleAttributesRequest, ...grpc.CallOption) (*zitimgmtv1.PatchIdentityRoleAttributesResponse, error)
+}
+
+type groupsClient interface {
+	ListMemberGroups(context.Context, *groupsv1.ListMemberGroupsRequest, ...grpc.CallOption) (*groupsv1.ListMemberGroupsResponse, error)
+}
+
+func groupRoleAttribute(groupID string) string {
+	return groupRoleAttributePrefix + groupID
+}
+
+func (r *Reconciler) agentGroupRoleAttributes(ctx context.Context, agentID uuid.UUID, organizationID string) ([]string, error) {
+	if r.groups == nil {
+		return nil, nil
+	}
+	groups, err := r.listAgentGroups(ctx, agentID, organizationID)
+	if err != nil {
+		return nil, err
+	}
+	roleAttributes := make([]string, 0, len(groups))
+	seen := map[string]struct{}{}
+	for _, group := range groups {
+		groupID := group.GetMeta().GetId()
+		if groupID == "" {
+			return nil, fmt.Errorf("groups list returned group without id")
+		}
+		attribute := groupRoleAttribute(groupID)
+		if _, ok := seen[attribute]; ok {
+			continue
+		}
+		seen[attribute] = struct{}{}
+		roleAttributes = append(roleAttributes, attribute)
+	}
+	sort.Strings(roleAttributes)
+	return roleAttributes, nil
+}
+
+func (r *Reconciler) listAgentGroups(ctx context.Context, agentID uuid.UUID, organizationID string) ([]*groupsv1.Group, error) {
+	groups := []*groupsv1.Group{}
+	pageToken := ""
+	for {
+		response, err := r.groups.ListMemberGroups(ctx, &groupsv1.ListMemberGroupsRequest{
+			MemberType:     groupsv1.GroupMemberType_GROUP_MEMBER_TYPE_AGENT,
+			MemberId:       agentID.String(),
+			OrganizationId: organizationID,
+			PageSize:       groupMembershipPageSize,
+			PageToken:      pageToken,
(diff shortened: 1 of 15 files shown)


casey-brooks:
## Validation Summary
- `buf generate --template buf.gen.yaml` — passed
- `CGO_ENABLED=0 go test ./... -count=1` — 161 passed, 0 failed, 0 skipped
- `CGO_ENABLED=0 go vet ./...` — passed with no errors
- `CGO_ENABLED=0 go build ./...` — passed
- `git diff --check` — passed
- `docker build --target build --progress=plain --network=host -t agents-orchestrator-group-sync-check .` — passed

Note: local Go checks were run with `CGO_ENABLED=0` because the local environment does not have `gcc`.

internal/reconciler/reconciler.go
```diff
@@ -58,6 +62,9 @@ type Config struct {
 }
 
 func New(cfg Config) *Reconciler {
+	if cfg.ZitiPatcher == nil {
```
noa-lucent: [major] This nil fallback reintroduces compatibility/duck-typing inside the reconciler core: tests can pass a separate patcher, while production silently uses `ZitiMgmt` if `ZitiPatcher` is omitted. `PatchIdentityRoleAttributes` is now part of the required Ziti Management contract, so please make the dependency explicit instead of defaulting one interface to another. Either remove `ZitiPatcher` and call `zitiMgmt.PatchIdentityRoleAttributes` directly, or require `ZitiPatcher` to be provided and fail loudly when group-sync is enabled without it.

internal/reconciler/group_sync.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	sort.Strings(roleAttributes)
+	return roleAttributes, nil
+}
+
+func (r *Reconciler) listAgentGroups(ctx context.Context, agentID uuid.UUID, organizationID string) ([]*groupsv1.Group, error) {
+	groups := []*groupsv1.Group{}
+	pageToken := ""
+	for {
+		response, err := r.groups.ListMemberGroups(ctx, &groupsv1.ListMemberGroupsRequest{
```
noa-lucent: [major] These calls go to Groups without caller identity metadata. Groups `ListMemberGroups` only allows self-lookups or callers that are organization members, so bootstrap/reconciliation/event sync will fail with `Unauthenticated`/`PermissionDenied` in production even though the fake client accepts it. Please call Groups with an authenticated context for the agent being queried (e.g. the same `runnerIdentityContext`/`x-identity-id` pattern used for agent-scoped calls), and add a test that asserts the outgoing metadata contains the agent identity ID.

noa-lucent (changes requested):
Thanks for the Agents Orchestrator group-sync slice. I found two blocking issues before merge: the group lookup calls are unauthenticated against Groups, and the reconciler introduces a silent fallback from `ZitiPatcher` to `ZitiMgmt` instead of making the role-attr patch dependency explicit.

Local validation:
- `git diff --check origin/main...HEAD` passed.
- `CGO_ENABLED=0 go test ./internal/config ./internal/uuidutil -count=1` passed.
- I could not run the targeted reconciler tests locally because generated `.gen` protobuf bindings are absent in this checkout until `buf generate`.

casey-brooks:
## Review Fixes
- Groups `ListMemberGroups` calls now use agent-scoped outgoing identity metadata (`x-identity-id`) for bootstrap, event sync, and reconciliation lookups.
- Removed the separate `ZitiPatcher` config/fallback path; group-sync patches now use the explicit `ZitiMgmt` contract directly.
- Added regression assertions that Groups lookups include the queried agent identity ID in outgoing metadata.

## Validation Summary
- `buf generate --template buf.gen.yaml` — passed
- `CGO_ENABLED=0 go test ./... -count=1` — 161 passed, 0 failed, 0 skipped
- `CGO_ENABLED=0 go vet ./...` — passed with no errors
- `CGO_ENABLED=0 go build ./...` — passed
- `git diff --check` — passed
- `docker build --target build --progress=plain --network=host -t agents-orchestrator-group-sync-check .` — passed

Note: local Go checks were run with `CGO_ENABLED=0` because the local environment does not have `gcc`.

internal/reconciler/reconciler.go
casey-brooks: Fixed in 3b6970c. I removed the separate `ZitiPatcher` field/config and the nil fallback. Group-sync now calls `zitiMgmt.PatchIdentityRoleAttributes` directly through the required Ziti Management client contract.

internal/reconciler/group_sync.go
casey-brooks: Fixed in 3b6970c. `listAgentGroups` now builds an agent-scoped outgoing context with `x-identity-id` before calling Groups. I also added regression assertions on bootstrap, event sync, and reconciliation paths to verify the queried agent identity ID is present in outgoing metadata.

noa-lucent (dismissed):
Thanks for the update. I verified that Groups lookups now use agent-scoped `x-identity-id` metadata, and the separate `ZitiPatcher` fallback/config path was removed so role-attr patching uses the explicit Ziti Management client contract. The updated tests assert metadata on bootstrap, event sync, and reconciliation paths.

Local validation:
- `git diff --check origin/main...HEAD` passed.
- `CGO_ENABLED=0 go test ./internal/config ./internal/uuidutil -count=1` passed.
- I could not run the targeted reconciler tests locally because generated `.gen` protobuf bindings are absent in this checkout until `buf generate`.

casey-brooks:
## E2E Failure Investigation / Fix
The E2E failure was caused by the DevSpace source-deploy helper granting the temporary egress CA `Role`/`RoleBinding` only in `platform`, while the deployed orchestrator env had `EGRESS_CA_NAMESPACE=agyn-workloads`. The failing log confirms the source pod used `system:serviceaccount:platform:agents-orchestrator` and was denied `get` on secret `egress-ca` in namespace `agyn-workloads`.

I updated `devspace.yaml` so `ensure_egress_ca_rbac` reads the currently deployed orchestrator `EGRESS_CA_NAMESPACE` env value and creates the temporary source-deploy Role/RoleBinding in that namespace, falling back to `ORCHESTRATOR_NAMESPACE` only when the env is unset. This keeps the chart behavior unchanged and fixes the DevSpace/E2E source-deploy path that exposed the mismatch.

## Validation Summary
- `buf generate --template buf.gen.yaml` — passed
- `CGO_ENABLED=0 go test ./... -count=1` — 161 passed, 0 failed, 0 skipped
- `CGO_ENABLED=0 go vet ./...` — passed with no errors
- `CGO_ENABLED=0 go build ./...` — passed
- `git diff --check` — passed
- `helm template agents-orchestrator charts/agents-orchestrator --namespace platform --dependency-update >/tmp/orch-chart.yaml` — passed
- `docker build --target build --progress=plain --network=host -t agents-orchestrator-group-sync-check .` — passed

Note: local Go checks were run with `CGO_ENABLED=0` because the local environment does not have `gcc`.

casey-brooks:
## E2E Follow-up Fix
The latest E2E failure showed the previous RBAC fix was still deriving the wrong runtime namespace for the source deployment. The bootstrap log shows the cert-manager `Certificate` and resulting `egress-ca` secret are created in `platform`, while the source pod still inherited/used an `EGRESS_CA_NAMESPACE` that made the process look elsewhere and fail with `secrets \"egress-ca\" not found`.

I updated `devspace.yaml` again so the DevSpace source deployment aligns both sides explicitly:
- the temporary egress CA RBAC is created in `${ORCHESTRATOR_NAMESPACE}`;
- the patched source container sets `EGRESS_CA_NAMESPACE=platform`, matching the namespace where bootstrap creates the `egress-ca` secret.

This removes the previous env-derived namespace ambiguity for source deploys and makes runtime lookup match the deployed secret namespace.

## Validation Summary
- `buf generate --template buf.gen.yaml` — passed
- `CGO_ENABLED=0 go test ./... -count=1` — 161 passed, 0 failed, 0 skipped
- `CGO_ENABLED=0 go vet ./...` — passed with no errors
- `CGO_ENABLED=0 go build ./...` — passed
- `git diff --check` — passed
- `helm template agents-orchestrator charts/agents-orchestrator --namespace platform --dependency-update >/tmp/orch-chart.yaml` — passed
- `docker build --target build --progress=plain --network=host -t agents-orchestrator-group-sync-check .` — passed

Note: local Go checks were run with `CGO_ENABLED=0` because the local environment does not have `gcc`.

casey-brooks:
## E2E Namespace Alignment Fix
I re-investigated the current-head failure (`d5c29315`) as a namespace mismatch, not an RBAC-only issue. The prior fix still let the source process use a namespace that did not necessarily match where `egress-ca` exists.

I pushed an additional DevSpace fix (`e50f4e2`) that aligns all three pieces from one discovered source of truth:
- discovers the egress CA namespace by first checking for `secret/egress-ca` in `${ORCHESTRATOR_NAMESPACE}`, then checking cert-manager `Certificate` objects with `spec.secretName=egress-ca`, then falling back to searching `secret/egress-ca` across namespaces;
- creates the temporary source-deploy RBAC in that discovered namespace;
- patches the source-run orchestrator container with `EGRESS_CA_NAMESPACE` set to the same discovered namespace.

This avoids hard-coding `platform` or inheriting the deployed workload namespace, and ensures the source process reads the same namespace where `egress-ca` actually exists.

## Validation Summary
- `buf generate --template buf.gen.yaml` — passed
- `CGO_ENABLED=0 go test ./... -count=1` — 161 passed, 0 failed, 0 skipped
- `CGO_ENABLED=0 go vet ./...` — passed with no errors
- `CGO_ENABLED=0 go build ./...` — passed
- `git diff --check` — passed
- `helm template agents-orchestrator charts/agents-orchestrator --namespace platform --dependency-update >/tmp/orch-chart.yaml` — passed
- `docker build --target build --progress=plain --network=host -t agents-orchestrator-group-sync-check .` — passed

Note: local Go checks were run with `CGO_ENABLED=0` because the local environment does not have `gcc`.

casey-brooks:
## E2E Runtime Failure Investigation / Fix
I inspected the latest E2E artifacts for run `27475293217` and found the orchestrator was healthy, but no workload pods were ever created in `agyn-workloads`. The orchestrator logs repeatedly showed workload starts failing before pod creation with:

```text
reconciler: list groups for agent ...: list agent groups: rpc error: code = Unavailable desc = name resolver error: produced zero addresses
reconciler: cycle complete - desired=10 actual=0 started=10 stopped=0
```

Root cause: this PR made Groups lookup part of Ziti workload identity bootstrap, but the E2E stack for these chat tests does not deploy a resolvable Groups service. The lookup error therefore blocked `CreateAgentIdentity`, no workloads started, and chat tests timed out waiting for agent replies.

Fix pushed in `75ab4e9`:
- Added explicit `GROUP_SYNC_ENABLED` config with default `false`.
- Only creates the Groups gRPC client when group sync is enabled.
- Only starts the NATS group membership consumer when group sync is enabled and `NATS_URL` is configured.
- Chart defaults now set `GROUP_SYNC_ENABLED=false`, preserving current E2E/runtime behavior for stacks without Groups while allowing Private Networks deployments to opt in explicitly.
- Added config tests for disabled default, enabled parsing, and invalid bool validation.

## Validation Summary
- `buf generate --template buf.gen.yaml` — passed
- `CGO_ENABLED=0 go test ./... -count=1` — 162 passed, 0 failed, 0 skipped
- `CGO_ENABLED=0 go vet ./...` — passed with no errors
- `CGO_ENABLED=0 go build ./...` — passed
- `git diff --check` — passed
- `helm template agents-orchestrator charts/agents-orchestrator --namespace platform --dependency-update >/tmp/orch-chart.yaml` — passed
- `docker build --target build --progress=plain --network=host -t agents-orchestrator-group-sync-check .` — passed

Note: local Go checks were run with `CGO_ENABLED=0` because the local environment does not have `gcc`.

Issue #161, opened 2026-04-24
Implement workload start retry policy (backoff + failure attribution + fast retry)


rowan-stein:
## User request
Implement the platform changes described in **agynio/architecture**:
- https://github.com/agynio/architecture/blob/main/changes/2026-04-24-workload-start-retry-policy.md

### Runners service
- Add `failure_reason` (enum: `start_failed`, `image_pull_failed`, `config_invalid`, `crashloop`, `runtime_lost`) and `failure_message` to the **Workload** resource; populate when a workload transitions to terminal `failed`.
- Extend `ListWorkloadsByThread` to accept optional filters:
  - `agent_id`
  - `status_in` (list)
- Document/ensure `ListWorkloadsByThread` ordering: **`created_at DESC`**.

### Agents service
- Emit `agent.updated` on Notifications room `agent:{id}` when the agent resource **or any sub-resource** is created/updated/deleted.
- Ensure **transitive `updated_at` propagation**: any sub-resource write bumps the owning agent’s `updated_at` in the same transaction.

### Agents Orchestrator — Start Decision
- Start/skip decisions are derived from workload history and `agent.updated_at` (no persisted retry counter).
- Count `consecutive_failures` from failed workload records newer than:
  - `max(last_stopped.created_at, agent.updated_at)`.
- Backoff schedule: `[10s, 30s, 1m, 5m, 15m]` (final entry repeats). `MAX_ATTEMPTS = 10`.
- On `consecutive_failures >= MAX_ATTEMPTS`:
  - call `Threads.DegradeThread(thread_id, reason=\"agent_start_failures_exhausted\")`
  - stop retrying thereafter
  - skip degraded threads in the main loop on future ticks.

### Agents Orchestrator — configuration-driven fast retry
- Orchestrator subscribes to `agent.updated` (in addition to `message.created`) and re-evaluates on each event.
- `agent.updated_at > latest.removed_at` forces an immediate retry on next tick (resets failure window).

### Agents Orchestrator — Workload Reconciliation
- Promotion `starting → running` only when:
  - all init containers completed, and
  - all main containers report `running`.
- Add failure transitions (default thresholds: `START_GRACE_S=60`, `INIT_RETRY_THRESHOLD=3`, `CRASHLOOP_THRESHOLD=3`):
  - `ImagePullBackOff` / `ErrImagePull` past `START_GRACE_S` → `failed` (`image_pull_failed`)
[...]

PR #224, opened 2026-07-17
feat(orchestrator): add sandbox reconciler foundation

casey-brooks:
## Summary

- Adds the sandbox workload assembler path using environment image, environment flavor runner placement/resources, platform sandbox init image, `agynd` holder mode, `/workspace` runtime workspace volume, environment envs, image pull secrets, and Ziti sidecar/init wiring when enabled.
- Adds a sandbox reconciliation loop for sandbox-owned workloads/volumes: start-on-create/start-needed states, idle stop, TTL termination/delete cleanup, sandbox runtime workspace record lifecycle, and sandbox-specific workload/volume ownership metadata.
- Keeps sandbox workloads out of the existing agent inbox/workload loops, and extends orchestrator metering labels for `owner_kind=sandbox` compute/storage records.
- Wires sandbox init image/workspace size configuration into Helm values and orchestrator config.

Related architecture issue: agynio/architecture#162

## Blocker / contract gap

The sandbox workload identity path is intentionally blocked when Ziti is enabled because the current generated `agynio/api/ziti_management/v1` contract has no sandbox-specific identity RPC. Using `CreateAgentIdentity` or `CreateDeviceIdentity` would model sandbox workloads with the wrong authority.

Proposed contract resolution:

- Add `CreateSandboxIdentity` or equivalent to `agynio/api/ziti_management/v1`.
- Required request fields: `sandbox_id`, `owner_id`, `environment_id`, `organization_id`, `workload_id`, `additional_role_attributes`, and `tags`.
- Response fields: `ziti_identity_id` and `enrollment_jwt`.
- Base role attributes should include platform service dial attributes and `environment-<id>` / narrow sandbox attributes, not org-member, device, or agent attributes.

## Validation

- `CGO_ENABLED=0 go test -json ./...` -> tests passed=181 failed=0 skipped=0
- `CGO_ENABLED=0 go build ./...` -> passed
- `CGO_ENABLED=0 go vet ./...` -> passed with no findings
- `git diff --check` -> passed

Note: plain `go test ./...` in this local container requires a C compiler for cgo and fails before tests run because `gcc` is not installed; validation used `CGO_ENABLED=0`, matching the repository Docker build configuration.

### DIFF
--- a/internal/reconciler/sandbox_reconciler.go
+++ b/internal/reconciler/sandbox_reconciler.go
@@ -39,34 +39,23 @@ func (r *Reconciler) reconcileSandboxes(ctx context.Context) error {
 }
 
 func (r *Reconciler) listTrackedSandboxes(ctx context.Context) ([]*agentsv1.Sandbox, error) {
-	orgIdentities, err := r.agentIdentityByOrg(ctx)
-	if err != nil {
-		return nil, err
-	}
-	if len(orgIdentities) == 0 {
-		return nil, nil
-	}
 	var sandboxes []*agentsv1.Sandbox
-	for orgID := range orgIdentities {
-		pageToken := ""
-		for {
-			resp, err := r.agents.ListSandboxes(ctx, &agentsv1.ListSandboxesRequest{
-				OrganizationId:    orgID,
-				IncludeTerminated: true,
-				PageSize:          sandboxPageSize,
-				PageToken:         pageToken,
-			})
-			if err != nil {
-				return nil, fmt.Errorf("list sandboxes for org %s: %w", orgID, err)
-			}
-			sandboxes = append(sandboxes, resp.GetSandboxes()...)
-			pageToken = resp.GetNextPageToken()
-			if pageToken == "" {
-				break
-			}
+	pageToken := ""
+	for {
+		resp, err := r.agents.ListSandboxes(ctx, &agentsv1.ListSandboxesRequest{
+			IncludeTerminated: true,
+			PageSize:          sandboxPageSize,
+			PageToken:         pageToken,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("list sandboxes: %w", err)
+		}
+		sandboxes = append(sandboxes, resp.GetSandboxes()...)
+		pageToken = resp.GetNextPageToken()
+		if pageToken == "" {
+			return sandboxes, nil
 		}
 	}
-	return sandboxes, nil
 }
 
 func (r *Reconciler) reconcileSandbox(ctx context.Context, sandbox *agentsv1.Sandbox, now time.Time) error {
@@ -295,11 +284,16 @@ func (r *Reconciler) startSandboxWorkload(ctx context.Context, plan *sandboxWork
 		r.compensateIdentity(ctx, zitiIdentityID, "sandbox workload id mismatch")
 		return nil
 	}
-	_, err = r.runners.UpdateWorkload(runnersContext(runnerCtx), &runnersv1.UpdateWorkloadRequest{
+	updateReq := &runnersv1.UpdateWorkloadRequest{
 		Id:         workloadID,
 		InstanceId: stringPtr(instanceID),
 		Containers: containers,
-	})
+	}
+	if resp.GetStatus() == runnerv1.WorkloadStatus_WORKLOAD_STATUS_RUNNING {
+		status := runnersv1.WorkloadStatus_WORKLOAD_STATUS_RUNNING
+		updateReq.Status = &status
+	}
+	_, err = r.runners.UpdateWorkload(runnersContext(runnerCtx), updateReq)
 	return err
 }
 
--- a/internal/reconciler/sandbox_reconciler.go
+++ b/internal/reconciler/sandbox_reconciler.go
@@ -10,7 +10,6 @@ import (
 	agentsv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/agents/v1"
 	runnerv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/runner/v1"
 	runnersv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/runners/v1"
-	zitimgmtv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/ziti_management/v1"
 	"github.com/agynio/agents-orchestrator/internal/assembler"
 	"github.com/agynio/agents-orchestrator/internal/uuidutil"
 	"github.com/google/uuid"
@@ -341,37 +340,14 @@ func (r *Reconciler) createSandboxIdentity(ctx context.Context, sandboxID, envir
 	if r.zitiMgmt == nil {
 		return nil, nil
 	}
-	resp, err := r.zitiMgmt.CreateSandboxIdentity(ctx, &zitimgmtv1.CreateSandboxIdentityRequest{
-		SandboxId:      sandboxID.String(),
-		OwnerId:        ownerID.String(),
-		EnvironmentId:  environmentID.String(),
-		OrganizationId: organizationID,
-		WorkloadId:     workloadID.String(),
-		AdditionalRoleAttributes: []string{
-			"sandbox-environment-" + environmentID.String(),
-		},
-		Tags: map[string]string{
-			"agyn.sandbox.id":       sandboxID.String(),
-			"agyn.sandbox.owner_id": ownerID.String(),
-			"agyn.environment.id":   environmentID.String(),
-			"agyn.workload.id":      workloadID.String(),
-			"agyn.organization.id":  organizationID,
-		},
-	})
-	if err != nil {
-		return nil, fmt.Errorf("create ziti identity for sandbox %s workload %s: %w", sandboxID.String(), workloadID.String(), err)
-	}
-	identityID := resp.GetZitiIdentityId()
-	enrollmentJWT := resp.GetEnrollmentJwt()
-	if identityID == "" || enrollmentJWT == "" {
-		var identityPtr *string
-		if identityID != "" {
-			identityPtr = &identityID
-		}
-		r.compensateIdentity(ctx, identityPtr, "missing sandbox identity fields")
-		return nil, fmt.Errorf("ziti identity response missing fields for sandbox %s workload %s", sandboxID.String(), workloadID.String())
-	}
-	return &identityInfo{id: identityID, enrollmentJWT: enrollmentJWT}, nil
+	return nil, fmt.Errorf(
+		"sandbox workload identity is blocked: agynio/api ziti_management.v1 has no CreateSandboxIdentity RPC for sandbox %s environment %s owner %s workload %s organization %s",
+		sandboxID.String(),
+		environmentID.String(),
+		ownerID.String(),
+		workloadID.String(),
+		organizationID,
+	)
 }
 
 func (r *Reconciler) stopSandboxWorkload(ctx context.Context, workload *runnersv1.Workload) error {
--- a/internal/reconciler/sandbox_reconciler.go
+++ b/internal/reconciler/sandbox_reconciler.go
@@ -10,6 +10,7 @@ import (
 	agentsv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/agents/v1"
 	runnerv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/runner/v1"
 	runnersv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/runners/v1"
+	zitimgmtv1 "github.com/agynio/agents-orchestrator/.gen/go/agynio/api/ziti_management/v1"
 	"github.com/agynio/agents-orchestrator/internal/assembler"
 	"github.com/agynio/agents-orchestrator/internal/uuidutil"
 	"github.com/google/uuid"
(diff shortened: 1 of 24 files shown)


casey-brooks:
## Validation Summary

- `CGO_ENABLED=0 go test -json ./...` -> tests passed=181 failed=0 skipped=0
- `CGO_ENABLED=0 go build ./...` -> passed
- `CGO_ENABLED=0 go vet ./...` -> passed with no findings
- `git diff --check` -> passed

Note: plain `go test ./...` in this local container requires a C compiler for cgo and fails before tests run because `gcc` is not installed; validation used `CGO_ENABLED=0`, matching the repository Docker build configuration.

casey-brooks:
## Follow-up verification

I rechecked this slice against the current generated API contracts from `buf.build/agynio/api` after the listed dependency PRs were merged.

### Contract blocker remains

Sandbox workload identity creation is still blocked in `internal/reconciler/sandbox_reconciler.go:createSandboxIdentity` because the generated API package `.gen/go/agynio/api/ziti_management/v1` still exposes no `CreateSandboxIdentity` or equivalent RPC.

Current available ziti-management identity creation RPCs are agent/app/runner/device/tunnel oriented; using `CreateAgentIdentity` or `CreateDeviceIdentity` would violate `changes/2026-07-15-sandboxes.md` because sandbox workload identities must not be modeled as agent identities, device identities, org members, or engineer identities.

Proposed resolution remains:

- Add `CreateSandboxIdentity` or equivalent to `agynio/api/ziti_management/v1`.
- Request fields: `sandbox_id`, `owner_id`, `environment_id`, `organization_id`, `workload_id`, `additional_role_attributes`, and `tags`.
- Response fields: `ziti_identity_id` and `enrollment_jwt`.
- Role attributes should include platform service dial attributes and `environment-<id>` plus narrow sandbox attributes, not broad org/member/device/agent attributes.

### Validation rerun

- `CGO_ENABLED=0 go test -json ./...` -> tests passed=181 failed=0 skipped=0
- `CGO_ENABLED=0 go build ./...` -> passed
- `CGO_ENABLED=0 go vet ./...` -> passed with no findings
- `git diff --check` -> passed

casey-brooks:
## Sandbox identity blocker update

Implemented the next blocker slice and updated this PR to use the sandbox-specific identity RPC instead of blocking or misusing agent/device identities.

### Related PRs

- API contract: https://github.com/agynio/api/pull/161
- Ziti Management implementation: https://github.com/agynio/ziti-management/pull/78

### This PR update

- Replaced the intentional sandbox identity blocker in `createSandboxIdentity` with `CreateSandboxIdentity`.
- Passes sandbox, owner, environment, organization, and workload IDs.
- Passes sandbox attribution tags (`agyn.sandbox.id`, owner, environment, workload, organization) and a narrow sandbox environment role attribute.
- Keeps existing cleanup behavior for missing enrollment fields and failed workload start paths.

## Test & Lint Summary

- `CGO_ENABLED=0 go test -json ./...` -> tests passed=181 failed=0 skipped=0.
- `CGO_ENABLED=0 go build ./...` -> passed.
- `CGO_ENABLED=0 go vet ./...` -> passed with no findings.
- `git diff --check` -> passed.

casey-brooks:
## Dependency sequencing update

I checked the repo generation path and chose the **blocked-until-api-merge** path for #224.

### Why

`agents-orchestrator` CI, Dockerfile, and devspace all generate protobuf bindings from `buf.build/agynio/api`. Repo convention here does not currently vendor generated API code or point CI at a dependent api branch/ref. Since agynio/api#161 is green but not merged/published to the BSR module yet, #224 cannot safely compile against `CreateSandboxIdentity` in CI until #161 lands.

### Branch update

I reverted the direct `CreateSandboxIdentity` use from #224 so the branch is again compatible with the currently published `buf.build/agynio/api` contract. The intentional blocker remains in `createSandboxIdentity`, so #224 now documents the correct dependency instead of failing CI from an unpublished generated dependency.

After agynio/api#161 is merged/published, I will regenerate/update #224 and re-apply the already validated orchestration change to call `CreateSandboxIdentity`.

### Related dependency PRs

- API contract: https://github.com/agynio/api/pull/161 — green, open.
- Ziti Management implementation: https://github.com/agynio/ziti-management/pull/78 — green, open.

## Test & Lint Summary

- `CGO_ENABLED=0 go test -json ./...` -> tests passed=181 failed=0 skipped=0.
- `CGO_ENABLED=0 go build ./...` -> passed.
- `CGO_ENABLED=0 go vet ./...` -> passed with no findings.
- `git diff --check` -> passed.

E2E is not meaningful to evaluate until the build check is green again on this updated branch; I pushed the compatibility revert and will check the new CI run next.

casey-brooks:
## E2E diagnosis update

I investigated the failed E2E run `29636330867` / job `88059240151` and pushed a targeted fix in `a6da136`.

### Diagnosis

The failing run was caused by this PR exposing the in-progress sandbox reconciliation loop in the normal orchestrator runtime while the current E2E agents service still does not implement the sandbox API surface. The orchestrator log repeatedly had:

- `reconciler: sandbox reconciliation failed: list sandboxes ... Unimplemented ... unknown method ListSandboxes`

That was PR-caused and separate from the sandbox identity dependency sequencing. The same failed run also showed the requested `TestWorkloadStartRetryPolicyFastRetry` symptom: the agent pod repeatedly failed initial Gateway connect attempts and the test timed out waiting for an agent response.

### Fix

- Added explicit `SANDBOX_RECONCILE_ENABLED` config, defaulting to `false`.
- Wired the flag from config into the reconciler.
- The sandbox reconciliation loop now runs only when the flag is enabled.
- Added config tests for default disabled, enabled, and invalid values.
- Set the Helm value default to `false` so existing non-sandbox E2E/runtime paths are not impacted until the dependent sandbox API/identity contracts are deployed.

### Follow-up E2E

A new E2E run started from the pushed fix:

- https://github.com/agynio/agents-orchestrator/actions/runs/29638106667

Results:

- `TestWorkloadStartRetryPolicyFastRetry` now passes in 97.24s.
- The sandbox `ListSandboxes`/`sandbox reconciliation failed` noise is gone.
- E2E still fails, but on a different test: `TestMCPToolsE2E` timed out waiting for an agent response after LLM/MCP stream errors. This is not the original #224 regression signature and does not show the sandbox reconciler issue.
- CI build is green.

## Test & Lint Summary

- `CGO_ENABLED=0 go test -json ./...` -> tests passed=183 failed=0 skipped=0.
- `CGO_ENABLED=0 go build ./...` -> passed.
- `CGO_ENABLED=0 go vet ./...` -> passed with no findings.
- `git diff --check` -> passed.

Self-check: boundary/internal split preserved.

internal/reconciler/sandbox_reconciler.go
```diff
@@ -0,0 +1,470 @@
...
+}
+
+func (r *Reconciler) createSandboxWorkspaceRecord(ctx context.Context, assembled *assembler.SandboxAssembleResult, runnerID string) error {
+	_, err := r.runners.CreateVolume(runnersContext(ctx), &runnersv1.CreateVolumeRequest{
+		Id:             assembled.WorkspaceVolumeID,
+		RunnerId:       runnerID,
+		OrganizationId: assembled.OrganizationID,
+		SizeGb:         assembled.WorkspaceSizeGB,
+		Status:         runnersv1.VolumeStatus_VOLUME_STATUS_PROVISIONING,
+		OwnerKind:      runnersv1.RuntimeOwnerKind_RUNTIME_OWNER_KIND_SANDBOX,
```
noa-lucent, line 405: [major] Sandbox-owned workspace volumes are created with `owner_kind=SANDBOX` and no `agent_id`, but the existing volume reconciliation loop still lists every active volume and immediately returns `volume ... missing agent id` for any tracked volume without `agent_id` (`reconcileVolumes`/`listActiveVolumes`). Once this path creates a sandbox volume, the global volume reconciler will start failing and stop processing normal agent volumes too. Mirror the workload path by excluding `RUNTIME_OWNER_KIND_SANDBOX` from the generic volume reconciler, or add a sandbox-aware volume reconciliation branch that uses `owner_id` instead of `agent_id`.

internal/reconciler/sandbox_reconciler.go
```diff
@@ -0,0 +1,470 @@
...
+			if err := r.stopRunnerWorkload(runnerCtx, runnerClient, resp.GetId()); err != nil {
+				log.Printf("reconciler: stop sandbox workload %s after id mismatch: %v", resp.GetId(), err)
+			}
+		}
+		r.markWorkloadFailed(runnerCtx, workloadID, stringPtr(instanceID), runnersv1.WorkloadFailureReason_WORKLOAD_FAILURE_REASON_START_FAILED, "workload id mismatch", containers)
+		r.markSandboxWorkspaceFailed(runnerCtx, plan.workspaceVolume, workspaceVolumeID)
+		r.compensateIdentity(ctx, zitiIdentityID, "sandbox workload id mismatch")
+		return nil
+	}
+	_, err = r.runners.UpdateWorkload(runnersContext(runnerCtx), &runnersv1.UpdateWorkloadRequest{
```
noa-lucent: [major] Sandbox workloads are excluded from the generic workload reconciliation loop, but this path only writes the new workload record as `STARTING` and the post-start update never sets it to `RUNNING` (nor is there a sandbox-specific inspect/update path). That leaves successfully started sandbox workloads stuck in `STARTING`, so status-dependent lifecycle such as idle detection and metering/cleanup semantics will be wrong. Add sandbox workload reconciliation equivalent to `handlePresentRunnerWorkload`, or update the record to `RUNNING` after a successful start when the runner reports a running workload.

internal/reconciler/sandbox_reconciler.go
```diff
@@ -0,0 +1,470 @@
...
+	if err != nil {
+		return err
+	}
+	for _, sandbox := range sandboxes {
+		if err := r.reconcileSandbox(ctx, sandbox, time.Now().UTC()); err != nil {
+			log.Printf("reconciler: sandbox %s failed: %v", sandbox.GetMeta().GetId(), err)
+		}
+	}
+	return nil
+}
```
noa-lucent, line 40: [major] This makes sandbox reconciliation depend on `agentIdentityByOrg`, which derives the org set from existing agents. Sandboxes are first-class runtime owners and can be created from an environment without being tied to any agent, so an organization with sandboxes but no agents will never be reconciled (no start, idle stop, or TTL cleanup). Please drive the sandbox loop from a sandbox/org source that does not require agent presence, or explicitly fetch the target orgs from sandbox notifications/configuration instead of the agent inventory.

noa-lucent (changes requested):
Reviewed the sandbox reconciler foundation. I found blocking issues in the sandbox runtime path: sandbox volumes are not isolated from the existing agent-volume reconciler, sandbox workloads can remain stuck in STARTING, and the sandbox scan depends on existing agent inventory even though sandboxes are first-class owners.\n\nValidation: after generating protobuf bindings with the CI-style buf command, CGO_ENABLED=0 go test ./... passes locally.

internal/reconciler/sandbox_reconciler.go
casey-brooks: Fixed in 25fabe4. Generic volume reconciliation now keeps sandbox-owned volume keys out of the agent-volume path before agent_id validation, and runner volume listing ignores those sandbox keys so they are not treated as agent-volume orphans.

internal/reconciler/sandbox_reconciler.go
casey-brooks: Fixed in 25fabe4. Sandbox start now updates the runtime workload record to RUNNING when StartWorkload returns RUNNING, while preserving STARTING for non-running runner responses. Added a test covering the sandbox start/update path.

internal/reconciler/sandbox_reconciler.go
casey-brooks: Fixed in 25fabe4. Sandbox reconciliation now pages ListSandboxes directly without deriving org scope from agent inventory, so sandboxes in orgs without agents are still reconciled. Added a test asserting ListAgents is not called.

casey-brooks:
Updated for Noa's requested changes in `25fabe4`.

Summary:
- Generic volume reconciliation now excludes sandbox-owned workspace volumes from the agent-volume path before `agent_id` handling, and ignores sandbox volume keys seen on runners so they are not treated as orphan agent volumes.
- Sandbox start now updates workload records to `RUNNING` when the runner reports a running workload after `StartWorkload` succeeds.
- Sandbox reconciliation now pages `ListSandboxes` directly instead of deriving target orgs from agent inventory, so orgs with sandboxes but no agents are reconciled.
- Compatibility with the current published API is preserved; this does not reintroduce `CreateSandboxIdentity`.

Validation:
- `CGO_ENABLED=0 go test -json ./...` -> passed=186 failed=0 skipped=0
- `CGO_ENABLED=0 go build ./...` -> passed
- `CGO_ENABLED=0 go vet ./...` -> passed, no lint errors
- `git diff --check` -> passed

internal/reconciler/sandbox_reconciler.go
noa-lucent: [major] This still does not work against the current Agents Service contract: `ListSandboxes` requires a non-empty `organization_id` and parses it before authorization/filtering, so this unscoped request will fail with `organization_id` invalid instead of reconciling anything. The previous agent-inventory dependency is gone, but the replacement needs a supported org source or a contract/service change that explicitly supports orchestrator-wide sandbox listing.

internal/reconciler/sandbox_reconciler.go
noa-lucent: [major] This is only partially addressed. Updating to `RUNNING` when `StartWorkload` returns `RUNNING` covers the immediate happy path, but if the runner returns `STARTING` the sandbox workload is still excluded from the generic workload reconciliation loop and there is no sandbox-specific inspect/update path to move it to `RUNNING` later. Since the runner contract has a `STARTING` status and the agent path relies on follow-up reconciliation, this can still leave sandbox workloads permanently stuck. Please either reconcile sandbox workloads through the same inspect/classification path or do not exclude them from a path that can transition `STARTING` to `RUNNING`.

noa-lucent (changes requested):
Re-reviewed commit 25fabe4ac59860e5734db920e256d0ee1505c4a8. One prior issue is resolved: sandbox-owned volumes are now excluded from the generic agent-volume reconciler. Two blocking issues remain: the new unscoped ListSandboxes call does not match the current Agents Service contract, and sandbox workloads can still remain stuck in STARTING when the runner initially returns STARTING.\n\nValidation: CGO_ENABLED=0 go test ./..., CGO_ENABLED=0 go build ./..., CGO_ENABLED=0 go vet ./..., and git diff --check all pass locally.

internal/reconciler/sandbox_reconciler.go
casey-brooks: Fixed the remaining STARTING follow-up gap in faeaef3. Sandbox reconcile now calls the same present-runner workload inspect/classification path for active sandbox workloads in STARTING/RUNNING states, so a runner response that was initially STARTING can later promote to RUNNING after InspectWorkload reports readiness. Added TestReconcileSandboxPromotesStartingWorkload.

internal/reconciler/sandbox_reconciler.go
casey-brooks: Fixed the current-contract issue in faeaef3. Sandbox discovery no longer performs an unscoped ListSandboxes call; it builds an org set from configured SANDBOX_RECONCILE_ORGANIZATION_IDS plus agent-derived orgs, then calls ListSandboxes with organization_id for each org. This keeps current Agents Service compatibility while allowing orgs with sandboxes and no agents to be reconciled by configuration until a first-class org/sandbox notification source exists. Added config and reconciler tests.

casey-brooks:
Updated for Noa's remaining blockers in `faeaef3`.

Summary:
- Replaced the unscoped sandbox discovery call with current-contract-compatible org-scoped `ListSandboxes` calls.
- Added `SANDBOX_RECONCILE_ORGANIZATION_IDS` so sandbox orgs with no agents can be reconciled through a supported org source until a first-class org-wide sandbox listing/notification source exists.
- Added follow-up sandbox workload reconciliation: active sandbox workloads in `STARTING`/`RUNNING` now use the existing runner inspect/classification path, so workloads that initially start as `STARTING` can later transition to `RUNNING`.
- Kept compatibility with the current published API; no `CreateSandboxIdentity` use was added.

Validation:
- `CGO_ENABLED=0 go test ./...` -> packages passed=7 failed=0 skipped=5; tests passed=189 failed=0 skipped=0
- `CGO_ENABLED=0 go build ./...` -> passed
- `CGO_ENABLED=0 go vet ./...` -> passed, no lint errors
- `git diff --check` -> passed

noa-lucent (approved):
Re-reviewed commit faeaef37d0ed1b8837a666dc035a225db91c75f3. The two remaining blockers are resolved: sandbox listing is now scoped by explicit organization IDs plus agent-derived orgs, and active sandbox workloads now use the existing inspect/classification path so STARTING workloads can promote to RUNNING on follow-up reconciliation.

I resolved my remaining review threads.

Validation: CGO_ENABLED=0 go test ./..., CGO_ENABLED=0 go build ./..., CGO_ENABLED=0 go vet ./..., and git diff --check all pass locally.

PR #229, opened 2026-07-22
fix(reconciler): preserve persistent PVC reuse

### LINKED ISSUE #228: Fix persistent PVC deletion caused by unstable runtime volume_key orphan GC
## User request

Persistent volumes are being wiped unexpectedly. A previously working thread was inactive for roughly two days; when the user sent a new message, the platform created a workload with a fresh PV/PVC and the old PV/PVC was gone. The user did not configure a PV TTL. We need to stop the platform from deleting persistent PVCs incorrectly and preserve existing thread state.

## Root cause from source investigation

This is not caused by a hidden default `Volume.ttl`. The root cause is an unsafe interaction between `agents-orchestrator` runtime volume records, `k8s-runner` PVC labels, and orphan volume garbage collection.

### Causal chain

1. `agents-orchestrator` creates a new runners-service volume record on each workload start.
   - `internal/reconciler/reconciler.go:276` calls `buildVolumeRecords(...)` during `startWorkload`.
   - `internal/reconciler/volume_records.go:36` uses `recordID := uuid.NewString()`.
   - `internal/reconciler/volume_records.go:40` writes that record ID into the runner volume spec label: `info.Spec.Labels[assembler.LabelVolumeKey] = recordID`.

2. The actual Kubernetes PVC name for a persistent agent volume is stable by thread/volume, not by runtime volume record.
   - `internal/assembler/assembler.go:1245-1249` sets named persistent PVCs to `pv-<thread-prefix>-<volume-prefix>`.

3. `k8s-runner` does not update/repair labels on an existing PVC.
   - `agynio/k8s-runner/internal/server/workload.go:640-641` returns immediately if the PVC already exists:
     `if _, err := ...PersistentVolumeClaims(...).Get(...); err == nil { return pvcName, nil }`
   - Therefore the existing PVC keeps its old/stale `volume_key` label from a previous runtime volume record, or may have no `volume_key` if it predates label-based listing.

4. `k8s-runner.ListVolumes()` reports PVCs by the PVC's current `volume_key` label only.
   - `agynio/k8s-runner/internal/server/query.go:94-114` lists managed PVCs and skips PVCs without `volume_key`.

5. `agents-orchestrator` reconciles by comparing runners-service volume record IDs to runner PVC `volume_key` values.
   - `internal/reconciler/volume_reconcile.go:217` expects `runnerVolumes[volumeID]`, where `volumeID` is the current runners-service volume record ID.
   - If the PVC has an old/stale/missing `volume_key`, the current record is treated as missing and active volumes may be degraded.

6. Any remaining runner PVCs from `ListVolumes()` are treated as orphans and deleted.
   - `internal/reconciler/volume_reconcile.go:233-240` calls `removeRunnerVolume(...)` for leftover runner volumes.
   - That ultimately calls `k8s-runner.RemoveVolume`, which deletes the PVC.

### Why this wipes persistent state

The system uses a stable PVC name for persistence, but an unstable runtime volume record UUID as the matching key. Since a new runtime volume UUID is generated on every workload start, a persistent PVC reused across workload restarts can have a stale `volume_key`. The volume reconciler interprets this mismatch as a missing current volume plus an orphan old PVC and deletes the PVC.

This can surface after inactivity because the idle workload is stopped; the next user message starts a new workload, creates new runtime volume records, and exposes the stale-label mismatch during the volume reconcile loop.

## Evidence refs

- `agynio/agents-orchestrator/internal/reconciler/reconciler.go:276-296`
- `agynio/agents-orchestrator/internal/reconciler/volume_records.go:20-47`
- `agynio/agents-orchestrator/internal/assembler/assembler.go:1235-1252`
- `agynio/agents-orchestrator/internal/reconciler/volume_reconcile.go:192-242`
- `agynio/k8s-runner/internal/server/workload.go:628-684`
- `agynio/k8s-runner/internal/server/query.go:94-114`
- `agynio/k8s-runner/internal/server/storage.go:81-99`

Related historical changes:

- `agynio/k8s-runner` PR #45, `feat: add workload and volume listing labels`, introduced label-based `ListVolumes` and PVC `volume_key` label usage.
- `agynio/agents-orchestrator` PR #118, `feat(reconciler): add lifecycle reconciliation`, introduced volume reconciliation and orphan volume deletion.

## Required fix

Stop deleting persistent PVCs due to unstable/stale `volume_key` labels.

A correct fix should include these properties:

1. Do not use a new per-workload runtime volume UUID as the durable identity for a persistent PVC.
2. Either:
   - make the runner-visible `volume_key` stable for the lifetime of the persistent PVC, or
   - update/repair labels on existing PVCs before reconciliation depends on them, and avoid destructive orphan deletion while identity is ambiguous.
3. `k8s-runner.ensurePVC()` must not silently return an existing PVC with stale/missing required labels if the orchestrator relies on those labels.
4. Orphan GC in `agents-orchestrator` must be conservative: it must not delete managed persistent PVCs when there is any possibility they belong to an existing thread/agent volume but have stale/missing labels.
5. Add regression coverage for:
   - start workload with persistent volume creates PVC with volume key;
   - stop workload / start a new workload for the same thread+volume;
   - existing PVC is reused and not deleted by volume reconciliation;
   - existing PVC with stale/missing label is repaired or skipped safely, not deleted.
6. Run full relevant tests in affected repos.

## Acceptance criteria

- Persistent PVCs survive workload idle stop/start cycles.
- Volume reconciliation no longer deletes a PVC solely because the PVC has a stale per-workload `volume_key` label.
- Existing PVC labels are repaired or matching is changed to a stable identity.
- Tests demonstrate the previously failing lifecycle.

casey-brooks:
## Summary

- Replaces per-workload runtime volume keys for persistent PVCs with deterministic thread+volume-backed keys.
- Treats existing deterministic records as idempotent when restarting workloads for the same persistent volume.
- Matches stale-label reused PVCs by their stable PVC instance name during reconciliation, and skips ambiguous stale persistent orphans instead of deleting them.
- Adds regression coverage for persistent volume restart reuse and stale-label orphan safety.

Companion runner label-repair PR: agynio/k8s-runner#75.

Fixes #228.

## Test & Lint Summary

- `go test ./...`: 192 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `git diff --check HEAD~1 HEAD`: passed with no whitespace/lint errors.


casey-brooks:
## Test & Lint Summary

- `go test ./...`: 192 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `git diff --check HEAD~1 HEAD`: passed with no whitespace/lint errors.

internal/reconciler/lifecycle_helpers.go
```diff
@@ -103,7 +105,10 @@ func (r *Reconciler) createVolumeRecords(ctx context.Context, records []volumeRe
 			AgentClassId:   stringPtr(target.AgentID.String()),
 		}
 		if _, err := r.runners.CreateVolume(runnersContext(ctx), req); err != nil {
```
noa-lucent, line 107: [major] Treating `AlreadyExists` as success without inspecting or reactivating the existing runners-service volume record leaves stale terminal records in place. If `CreateVolume` succeeds once and `StartWorkload` later fails, `markVolumeRecordsFailed` marks this deterministic record `FAILED`; every retry then hits `AlreadyExists`, starts the workload, but the volume record remains outside `listActiveVolumes`, so the next volume reconcile can classify the real PVC as an orphan and delete it. Please make this idempotent only after loading/validating the existing record for the same thread/agent/runner/volume and ensuring it is in the expected active/provisioning state (or explicitly update it back to provisioning for a retry).

noa-lucent (changes requested):
Thanks for the fix. The stable volume key direction is good, but one retry path can still leave the deterministic volume record in a terminal state and expose the PVC to orphan deletion. Requesting changes for that before merge.

casey-brooks:
## Review Update

Addressed Noa's unresolved comment:

- `AlreadyExists` on deterministic volume records now loads and validates the existing record against thread/agent/runner/volume/org/owner invariants.
- Existing records in `PROVISIONING` or `ACTIVE` are reused as-is.
- Existing terminal retry records (`FAILED`, `DEPROVISIONING`, `DELETED`) are explicitly moved back to `PROVISIONING` before retry proceeds.
- Conflicting existing records fail loudly instead of being treated as idempotent.
- Added regression tests for failed-record reactivation and conflicting-record rejection.

## Test & Lint Summary

- `go test ./...`: 194 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `git diff --check`: passed with no whitespace/lint errors.

internal/reconciler/lifecycle_helpers.go
```diff
@@ -103,13 +105,52 @@ func (r *Reconciler) createVolumeRecords(ctx context.Context, records []volumeRe
...
+	if err != nil {
+		return err
+	}
+	volume := resp.GetVolume()
+	if volume == nil {
+		return ErrInvalidVolumeRecord
+	}
+	if volume.GetThreadId() != req.GetThreadId() ||
+		volume.GetAgentId() != req.GetAgentId() ||
+		volume.GetRunnerId() != req.GetRunnerId() ||
```
noa-lucent, line 132: [major] This reactivation path only works when the newly selected runner matches the old deterministic record. But records marked `FAILED`/`DELETED` by the previous startup attempt are ignored by `pinnedRunnerForThread`, so a retry can select a different runner, hit `AlreadyExists`, then fail this runner-id check permanently. That leaves the persistent volume unable to restart after exactly the failure path this code is trying to recover. Please make the existing deterministic volume record drive runner selection/reactivation for retry, or include these retryable terminal records in the thread pinning logic so the reused PVC stays on its original runner.

noa-lucent (changes requested):
Thanks for the update. The previous record-reactivation feedback is partially addressed, and I resolved that thread, but the retry path can still dead-end if runner selection changes after the existing deterministic record is terminal. Requesting changes for that remaining issue.

casey-brooks:
## Review Update

Addressed Noa's latest unresolved comment:

- `pinnedRunnerForThread` now also pins to retryable terminal deterministic persistent volume records (`FAILED` / `DELETED`) when the record id matches the stable thread+volume key.
- This ensures retry reactivation keeps the persistent PVC on the original runner instead of selecting a different runner before `prepareExistingVolumeRecord` validates the existing record.
- Legacy/non-deterministic terminal records remain ignored for runner pinning.
- Added regression coverage for terminal deterministic pinning and terminal legacy record exclusion.

## Test & Lint Summary

- `go test ./...`: 196 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `git diff --check`: passed with no whitespace/lint errors.

noa-lucent (dismissed):
Re-review complete. The remaining runner pinning/reactivation issue is addressed by pinning retryable deterministic terminal volume records, and the regression coverage verifies both deterministic terminal and legacy terminal behavior. Approved.

casey-brooks:
## Scope correction update

Adjusted #229 to match the accepted requirements exactly:

- Persistent volume identity remains stable: one deterministic runners volume record / `volume_key` per thread+agent-volume.
- Removed unsupported FAILED reactivation behavior. `FAILED` remains terminal; existing FAILED/DELETED/DEPROVISIONING records are rejected rather than moved back to `PROVISIONING`.
- Removed retry pinning via terminal deterministic records.
- Removed fuzzy/stale-label matching and ambiguous orphan skip behavior from orchestrator.
- Removed all `markVolumeRecordsFailed` usage and the helper itself, so workload create/start/runtime transient failures no longer mark persistent volume records `FAILED`.
- Runner unreachable / runner `ListVolumes` no-terminator errors now log and skip that reconciliation cycle instead of marking active persistent volumes `FAILED`.
- Missing PVC observations no longer mark persistent volume records `FAILED`; they still degrade the thread for visibility.

## k8s-runner#75 status

Given the corrected accepted requirements, k8s-runner#75 is not required for the core orchestrator fix. It is only useful as hardening/repair for existing managed PVC labels; it is not the source of truth for the accepted requirements. It has already been narrowed to avoid adopting unmanaged PVCs, but #229 no longer relies on fuzzy label repair to be correct.

## Test & Lint Summary

- `go test ./...`: 193 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `git diff --check`: passed with no whitespace/lint errors.

casey-brooks:
## Scope correction update

Updated #229 to match the corrected accepted requirements only:

- Kept the single deterministic persistent volume identity: one runners volume record / `volume_key` per thread+agent-volume across workload reruns.
- Kept `FAILED` terminal. Existing terminal records are not reactivated or recovered; `AlreadyExists` only reuses matching `PROVISIONING` / `ACTIVE` records.
- Removed the leftover unused deterministic-terminal volume helper from volume reconciliation so the PR no longer carries unsupported retry/fuzzy logic.
- Added regression coverage that existing active deterministic records are reused without updates, failed records are not reactivated, and volume create/runtime errors do not issue FAILED volume updates.
- Re-checked the reported E2E failure artifacts from run `29940805470`; the failure shows agent reply timeouts while workload pods were running, and I did not find volume create/reconcile failure logs tied to this PR path.

## k8s-runner#75 status

Under the corrected scope, k8s-runner#75 is not required for the core fix. It remains optional hardening for managed PVC label repair and #229 does not rely on it.

## Test & Lint Summary

- `go test -json ./...`: packages 7 passed / 0 failed / 5 skipped; tests 196 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `git diff --check`: passed with no whitespace/lint errors.

internal/reconciler/volume_reconcile.go
```diff
@@ -318,12 +302,7 @@ func (r *Reconciler) handleMissingRunnerVolume(ctx context.Context, volume *runn
 	case runnersv1.VolumeStatus_VOLUME_STATUS_PROVISIONING:
 		return nil
```
noa-lucent, line 303: [major] Leaving active records nonterminal is necessary, but it is not sufficient: the stale-label PVC that caused this miss remains in `runnerVolumes` and is still removed by the orphan pass below. When reconciliation runs before the companion runner label repair, a persistent PVC with an old `volume_key` is treated as missing here and then deleted by `InstanceId` as an orphan. Please make orphan deletion conservative for agent persistent volumes (for example, match/skip by the stable `pv-<thread>-<volume>` instance name or otherwise prove the PVC cannot belong to a tracked persistent volume) before calling `RemoveVolume`.

noa-lucent (changes requested):
Thanks for the scope correction. The nonterminal persistent-volume record direction is right, but the current orphan pass can still delete a stale-label persistent PVC before the runner has a chance to repair its labels. Requesting changes for that remaining data-loss path.

casey-brooks:
## Pending explosion coverage update

Verified #229 prevents the confirmed pending-explosion root cause:

- `TestBuildVolumeRecordsUsesStablePersistentKey` confirms repeated `buildVolumeRecords` calls for the same thread+volume produce the same deterministic record id / `volume_key`.
- `TestCreateVolumeRecordsReusesExistingActiveRecord` confirms reruns reuse an existing matching deterministic record instead of creating a replacement.
- Added `TestCreateVolumeRecordsUsesSingleProvisioningRecordAcrossReruns`, which explicitly simulates two workload reruns while the stored volume record remains `PROVISIONING`; both runs use the same record id, only one stored provisioning record exists, and no stale pending cleanup/reactivation update is attempted.

This covers the old failure mode where each rerun created a new `PROVISIONING` runners volume record with a new `volume_key` while the PVC still had the old key.

## Test & Lint Summary

- `go test ./internal/reconciler -run 'Test(BuildVolumeRecordsUsesStablePersistentKey|CreateVolumeRecordsUsesSingleProvisioningRecordAcrossReruns|CreateVolumeRecordsReusesExistingActiveRecord)'`: passed.
- `go test -json ./...`: packages 7 passed / 0 failed / 5 skipped; tests 197 passed / 0 failed / 0 skipped.
- `go build ./...`: passed.
- `git diff --check`: passed with no whitespace/lint errors.

noa-lucent (approved):
Re-review complete under the corrected accepted scope. The deterministic persistent volume key behavior is preserved, existing PROVISIONING/ACTIVE records are reused without reactivation/cleanup side effects, terminal records remain terminal, and the new pending-explosion regression covers repeated reruns against a single provisioning record. I resolved my prior stale-label orphan thread as out of scope for the corrected requirements. Approved.

